PR #50, opened 2019-08-20
Fixed link for lerna

wreiske:
lernajs.io is not a working domain. Updated the README to show https://lerna.js.org/ instead

See: https://github.com/lerna/lerna#readme-badge


tassoevan:
Thanks a lot!

PR #113, opened 2019-12-16
feat: fuselage-datavis

tassoevan:


tassoevan:
I think nivo provides everything we need, so there is no need to use it behind a new fuselage package.

PR #154, opened 2020-03-04
perf: Avoid re-renders due to PropsContext

tassoevan:
Each `Box` consumes and provides an *new* plain object value from `PropsContext`, that causes pointless re-renderings. Using a single empty object to represent no additional props and component memoization might fix that.


tassoevan:
@gabriellsh Could you please test it working on Rocket.Chat? Comparing the latest version against this branch via React DevTools profilling would be enough.

gabriellsh:
> @gabriellsh Could you please test it working on Rocket.Chat? Comparing the latest version against this branch via React DevTools profilling would be enough.

tested, looks way better, thanks!

PR #155, opened 2020-03-04
feat: Added callback prop for scroll position

gabriellsh:
Scrollable now accepts scrollCb that receives `{ top: bool, bottom: bool, left: bool, right: bool}`.

The condition for each boolean is that content is hidden in that direction (eg. if scrolled down, top will be true).

### DIFF
--- a/packages/fuselage/src/components/Box/Scrollable/index.js
+++ b/packages/fuselage/src/components/Box/Scrollable/index.js
@@ -1,9 +1,24 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { useProps } from '../../../hooks';
 
-export function Scrollable({ children, horizontal, vertical, smooth }) {
+export function Scrollable({ children, horizontal, vertical, smooth, scrollCb }) {
+  let scrollTimeout;
+
+  const handleScroll = useCallback(function(event) {
+    const { target } = event;
+    if (!scrollTimeout) {
+      scrollCb({ top: !!target.scrollTop, bottom: !!(target.scrollTop + target.clientHeight - target.scrollHeight), left: !!target.scrollLeft, right: !!(target.scrollLeft + target.clientWidth - target.scrollWidth) });
+    } else {
+      clearTimeout(scrollTimeout);
+    }
+    scrollTimeout = setTimeout(() => {
+      scrollTimeout = false;
+      scrollCb({ top: !!target.scrollTop, bottom: !!(target.scrollTop + target.clientHeight - target.scrollHeight), left: !!target.scrollLeft, right: !!(target.scrollLeft + target.clientWidth - target.scrollWidth) });
+    }, 200);
+  }, [scrollCb]);
+
   const [, PropsProvider] = useProps(({ className, ...props }) => ({
     className: [
       className,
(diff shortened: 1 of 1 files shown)


packages/fuselage/src/components/Box/Scrollable/index.js
```diff
@@ -1,9 +1,24 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { useProps } from '../../../hooks';
 
-export function Scrollable({ children, horizontal, vertical, smooth }) {
+export function Scrollable({ children, horizontal, vertical, smooth, scrollCb }) {
```
tassoevan: Better replace `scrollCb` with something prefixed with `on-`. Maybe `onScrollToEdge`.

packages/fuselage/src/components/Box/Scrollable/index.js
```diff
@@ -1,9 +1,24 @@
...
+  let scrollTimeout;
```
tassoevan: Prefer to share variables across re-renders via `useRef`: declare with `const scrollTimeoutRef = useRef();` and use it as `scrollTimeoutRef.current`. Scopes of functional components should be as discardable as possible to avoid memory leaks; keep only on the hooked state.

packages/fuselage/src/components/Box/Scrollable/index.js
```diff
@@ -1,9 +1,24 @@
...
+    const { target } = event;
+    if (!scrollTimeout) {
+      scrollCb({ top: !!target.scrollTop, bottom: !!(target.scrollTop + target.clientHeight - target.scrollHeight), left: !!target.scrollLeft, right: !!(target.scrollLeft + target.clientWidth - target.scrollWidth) });
+    } else {
+      clearTimeout(scrollTimeout);
+    }
+    scrollTimeout = setTimeout(() => {
+      scrollTimeout = false;
+      scrollCb({ top: !!target.scrollTop, bottom: !!(target.scrollTop + target.clientHeight - target.scrollHeight), left: !!target.scrollLeft, right: !!(target.scrollLeft + target.clientWidth - target.scrollWidth) });
+    }, 200);
```
tassoevan: What about `scrollTimeout === undefined ? 0 : 200`? That if would become superfluous and some code duplication will vanish.

packages/fuselage/src/components/Box/Scrollable/index.js
```diff
@@ -22,4 +38,5 @@ Scrollable.propTypes = {
   horizontal: PropTypes.bool,
   vertical: PropTypes.bool,
   smooth: PropTypes.bool,
+  scrollCb: PropTypes.func,
```
tassoevan: Don't forget to update the field here.

packages/fuselage/src/components/Box/Scrollable/index.js
```diff
@@ -12,6 +27,7 @@ export function Scrollable({ children, horizontal, vertical, smooth }) {
       vertical && 'rcx-box--scrollable-vertical',
       smooth && 'rcx-box--scrollable-smooth',
     ].filter(Boolean).join(' '),
+    onScroll: typeof onScrollContent !== 'undefined' ? handleScroll : null,
```
tassoevan:
```suggestion
    onScroll: typeof onScrollContent !== 'undefined' ? handleScroll : undefined,
```

Issue #162, opened 2020-03-06
SelectFiltered: Wrong keyboard navigation


gabriellsh:
**Description**
Choosing an option with the arrow keys selects the wrong one when filter is applied.

**Steps to reproduce**
Open select, filter results, select option with keyboard and press enter.

**Insert an image/gif**

![Select keyboard filter bug](https://user-images.githubusercontent.com/40830821/76116268-72a90980-5fc8-11ea-81e3-05fb930a4877.gif)

TaimurAzhar:
Hello, I think I have solved this issue. Creating the PR currently. It took me a while.

dougfabris:
Fixed here: #169

PR #169, opened 2020-03-15
fix: SelectFiltered: Wrong keyboard navigation

TaimurAzhar:
Changed the options object that was sent to the _Options in the Select.js file. Please let me know if the issues need to be resolved in any other places as well and I will do so accordingly.

fixes issue #162 

smallGif


TaimurAzhar:
Thanks for reviewing.

PR #187, opened 2020-04-01
fix: Padding props

tassoevan:
Padding value were ignored.


packages/fuselage/src/components/Tile/index.js
```diff
@@ -7,13 +7,11 @@ import { Box } from '../Box';
 export const Tile = forwardRef(function Tile({
   className,
   elevation,
-  padding,
   ...props
 }, ref) {
   return <Box
     ref={ref}
     componentClassName='rcx-tile'
-    p={padding}
```
tassoevan, line 16: Now the `padding` prop is processed by `Box` itself.

PR #188, opened 2020-04-01
feat: Updated icons

tassoevan:
Updates our icon list. Additionally, now we:

- keep a `glyphsMapping.json` file that guarantees each icon will be uniquely used by the same unicode character across future versions;
- fix broken symbolic links for aliases;
- prefer naming icons according to what they look instead of what feature they represent.


packages/icons/src/neutral/home.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
tassoevan, line 1: Keep only the `xmlns` and the `viewBox` attributes, since the other ones are irrelevant.

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32">
```

packages/icons/src/neutral/home.svg
```diff
@@ -0,0 +1,3 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M18.0714 5.59483C16.9121 4.48825 15.0878 4.48825 13.9285 5.59483L7.3095 11.913L4.3095 14.7766C4.01386 15.0588 3.91971 15.4925 4.07171 15.8719C4.22371 16.2513 4.59128 16.5 4.99997 16.5H6.99997V25C6.99997 26.6569 8.34312 28 9.99997 28H22C23.6568 28 25 26.6569 25 25V16.5H27C27.4087 16.5 27.7762 16.2513 27.9282 15.8719C28.0802 15.4925 27.9861 15.0588 27.6904 14.7766L24.6904 11.913L18.0714 5.59483ZM8.99997 15.5V13.0643L15.3095 7.04154C15.6959 6.67268 16.304 6.67268 16.6904 7.04154L23 13.0643V25C23 25.5523 22.5523 26 22 26H18V20C18 19.4477 17.5523 19 17 19H15C14.4477 19 14 19.4477 14 20V26H9.99997C9.44769 26 8.99997 25.5523 8.99997 25V15.5Z" fill="#2F343D"/>
```
tassoevan, line 2: No filling attribute should be kept, unless the path breaks.

```suggestion
<path d="M18.0714 5.59483C16.9121 4.48825 15.0878 4.48825 13.9285 5.59483L7.3095 11.913L4.3095 14.7766C4.01386 15.0588 3.91971 15.4925 4.07171 15.8719C4.22371 16.2513 4.59128 16.5 4.99997 16.5H6.99997V25C6.99997 26.6569 8.34312 28 9.99997 28H22C23.6568 28 25 26.6569 25 25V16.5H27C27.4087 16.5 27.7762 16.2513 27.9282 15.8719C28.0802 15.4925 27.9861 15.0588 27.6904 14.7766L24.6904 11.913L18.0714 5.59483ZM8.99997 15.5V13.0643L15.3095 7.04154C15.6959 6.67268 16.304 6.67268 16.6904 7.04154L23 13.0643V25C23 25.5523 22.5523 26 22 26H18V20C18 19.4477 17.5523 19 17 19H15C14.4477 19 14 19.4477 14 20V26H9.99997C9.44769 26 8.99997 25.5523 8.99997 25V15.5Z" />
```

tassoevan:
@gabriellsh
- I've renamed `add-reaction` to `emoji-plus`, and others as well.
- I'll apply most of the suggestioms, but `mobile` exists as a mirror to `desktop`; should it be renamed too?
- The duplicates exist because they're temporary aliases to older icon names that we need to deprecate on RC first.

gabriellsh:
It's okay for me. I just mentioned `mobile` because the word itself is an adjective, but `Desktop` isn't. Maybe it could be `mobile-phone`, but I think there's no problem in having an odd one out.

PR #222, opened 2020-04-30
perf!: Replace PropsContext

tassoevan:


tassoevan:
Until a good solution for the problem of replacing props that add classNames -- only achievable through `PropsProvider` -- emerge, individual steps of this PR were converted to their own PRs.

PR #236, opened 2020-06-02
chore: Live playground on Storybook

gabriellsh:
With ability to choose code highlight theme and preview position (side or bottom)

image

image

image


packages/fuselage/src/playground/index.js
```diff
@@ -0,0 +1,99 @@
...
+import shadesOfPurple from 'prism-react-renderer/themes/shadesOfPurple';
+import synthwave84 from 'prism-react-renderer/themes/synthwave84';
+import ultramin from 'prism-react-renderer/themes/ultramin';
+import vsDark from 'prism-react-renderer/themes/vsDark';
+import * as cssInJs from '@rocket.chat/css-in-js';
+import * as hooks from '@rocket.chat/fuselage-hooks';
+
+import * as components from '..';
+
+const scope = { components, React, cssInJs, hooks };
```
tassoevan, line 25: Why not spread all the symbols? I don't think they collide:

```suggestion
const scope = { ...components, ...React, ...cssInJs, ...hooks };
```

packages/fuselage/src/playground/index.js
```diff
@@ -0,0 +1,99 @@
...
+import duotoneLight from 'prism-react-renderer/themes/duotoneLight';
+import github from 'prism-react-renderer/themes/github';
+import nightOwl from 'prism-react-renderer/themes/nightOwl';
+import nightOwlLight from 'prism-react-renderer/themes/nightOwlLight';
+import oceanicNext from 'prism-react-renderer/themes/oceanicNext';
+import palenight from 'prism-react-renderer/themes/palenight';
+import shadesOfPurple from 'prism-react-renderer/themes/shadesOfPurple';
+import synthwave84 from 'prism-react-renderer/themes/synthwave84';
+import ultramin from 'prism-react-renderer/themes/ultramin';
+import vsDark from 'prism-react-renderer/themes/vsDark';
```
tassoevan, line 19: Pick just one.

gabriellsh: Sorry if I believe in a world where programmers can express themselves using whatever theme they want...

Jk, I guess I just went a little too overboard with it 😝 .

tassoevan: To be fair we can rollback when Storybook allow us to switch light and dark mode in the future.

PR #243, opened 2020-06-19
fix: CheckBox and RadioButton style when parent direction is RTL

sadeghpro:
Check box and radio button have a bug in rtl page. for exapmle radio button in rocket chat with rtl languages like Persian or Arabic:
Screen Shot 2020-06-19 at 6 53 05 PM
or check box :
Screen Shot 2020-06-19 at 6 53 52 PM


sadeghpro:
Hi @tassoevan 
Do I have to do anything?

tassoevan:
I've checked this PR at least three times and I never was able to reproduce the issue it is supposed to fix, nor have found any issue about it. I'm closing it by now, but I will gladly reopen it if more details are going to be described.

sadeghpro:
tnx @tassoevan yes it's fixed

PR #315, opened 2020-11-05
feat: Remove Box dependence from Tag and Badge components

tiagoevanp:


packages/fuselage/src/components/Tag/styles.scss
```diff
@@ -71,15 +71,11 @@ $tag-colors-disabled-background-color: theme('tag-colors-disabled-background-col
...
-    overflow: hidden;
-
-    text-align: center;
-
-    border-radius: lengths.border-radius(full);
+  &--small {
+    @include typography.use-font-scale(c2);
   }
 
-  &--clickable {
```
gabriellsh, line 82: Was the clickable modifier/prop supposed to be removed?

packages/fuselage/src/components/Badge/styles.scss
```diff
@@ -0,0 +1,82 @@
...
+    color: $badge-colors-ghost-color;
+    background-color: $badge-colors-ghost-background-color;
+  }
+
+  &--disabled {
+    color: $badge-colors-disabled-color;
+    background-color: $badge-colors-disabled-background-color;
+  }
+
+  &--round {
```
gabriellsh, line 75: I don't think this is necessary anymore. You can apply this style directly to the component. I had used the `rounded` prop because I was using the same component for tags and badges. But badges are always round.

gabriellsh (changes requested):

ggazzo (approved):

PR #330, opened 2020-11-13
fix: Set a conservative output.environment on Webpack bundles

tassoevan:
It should get rid of ES2015 syntax on Webpack support code in bundles.


MartinSchoeler (approved):
Looks good to me, this fixes the problems with IE and it doesn't seem to disrupt older versions of RC as well (3.7.x)

PR #337, opened 2020-11-30
feat: useClipboard hook

dougfabris:


packages/fuselage-hooks/src/useClipboard.ts
```diff
@@ -0,0 +1,57 @@
...
+  onCopySuccess?: (e?: Event) => void;
+  onCopyError?: (e?: Event) => void;
+};
+
+export type UseClipboardReturn = {
+  copy: (e?: Event) => Promise<void>;
+  hasCopied: boolean;
+};
+
+export const useClipboard = (
```
tassoevan, line 23:
```suggestion
type UseClipboardParams = {
  clearTime?: number;
  onCopySuccess?: (e?: Event) => void;
  onCopyError?: (e?: Event) => void;
};

export type UseClipboardReturn = {
  copy: (e?: Event) => Promise<void>;
  hasCopied: boolean;
};

/**
 * Hook to copy the passed content to the clipboard.
 *
 * @returns an object with the copy function and the hasCopied state
 * @public
 */
export const useClipboard = (
```

Doc comments have the inconvenient of requiring to be placed above the declaration they refer, without blank lines between them.

packages/fuselage-hooks/README.md
```diff
@@ -101,6 +103,20 @@ Hook to catch which responsive design' breakpoints are active.
 
 Returns **[Array](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)&lt;[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)>** an array of the active breakpoint names
 
+### UseClipboardParams
+
+Hook to copy the passed content to the clipboard.
```
tassoevan, line 108: This is wrong due to a misplaced doc comment.

tassoevan (changes requested):

packages/fuselage-hooks/README.md
dougfabris: New docs generated

packages/fuselage-hooks/src/useClipboard.ts
dougfabris: Nice to know about. I already fixed it

tassoevan (approved):

PR #372, opened 2021-01-25
[FIX] Remove padding for tags

### LINKED ISSUE #255: Tag: Padding
Padding block for Tag, should be 0 instead of 2px... :/

im-adithya:
This PR closes #255


im-adithya:
@ggazzo is this what you wanted? Or should I also add a padding prop to set required padding?

dougfabris:
@im-adithya it's already fixed here: #437

PR #560, opened 2021-10-22
test(fuselage,fuselage-hooks): Concise tests

tassoevan:
- [x] I have signed the CLA
- [ ] I have run Loki's visual regression tests (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)

- Fixes some lint configuration issues;
- Adds a `testing-utils` too, for Jest setup and mocking;
- Uses `@testing-library/react`, `@testing-library/jest-dom`, and `@testing-library/react-hooks` to make tests less verbose and more descriptive and meaning;


packages/fuselage-hooks/src/useDebouncedUpdates.spec.ts
```diff
@@ -1,50 +1,38 @@
-import { createElement, useState, FunctionComponent, StrictMode } from 'react';
-import { render } from 'react-dom';
-import { act } from 'react-dom/test-utils';
+import { renderHook, act } from '@testing-library/react-hooks';
+import { useState } from 'react';
 
-import { useDebouncedUpdates } from '.';
+import { useDebouncedUpdates } from './useDebouncedUpdates';
 
-describe('useDebouncedUpdates hook', () => {
```
dougfabris, line 7: We won't use describe anymore?

dougfabris, line 7: We won't use `describe` anymore?

packages/fuselage-hooks/src/useLocalStorage.server.spec.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
dougfabris, line 1: Why we're using this `.server` tests for this kind of functionality only present on the client side?

tools/testing-utils/lazySingletonStyleTagModule.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
dougfabris, line 1: why not `.ts`?

packages/fuselage/src/components/Table/index.d.ts
```diff
@@ -19,6 +20,7 @@ type TableCellProps = ComponentProps<typeof Box> & {
 export const Table: ForwardRefExoticComponent<TableProps> & {
   Head: ForwardRefExoticComponent<TableHeadProps>;
   Body: ForwardRefExoticComponent<TableBodyProps>;
+  Foot: ForwardRefExoticComponent<TableFootProps>;
```
dougfabris, line 23: `Footer` isn't a better name here?

packages/fuselage-hooks/src/useLocalStorage.server.spec.ts
ggazzo: ssr

packages/fuselage-hooks/src/useDebouncedUpdates.spec.ts
tassoevan: Jest doesn't require it, as the test suite filename itself works to name the test suite -- it goes into the cli output regardless. Better keep using `describe` only to limit special conditions for tests, like sections that require `before` and `after` hooks.

packages/fuselage-hooks/src/useLocalStorage.server.spec.ts
tassoevan: Yeah, SSR. A React component using our hooks should be able to run on Node.js to generate static HTML, therefore we must guarantee that they don't break when web APIs are missing.

tools/testing-utils/lazySingletonStyleTagModule.js
tassoevan: I couldn't import a `*.ts` file directly from a package, only `*.js` with sibling `*.d.ts` files.

packages/fuselage/src/components/Table/index.d.ts
tassoevan: Maybe. I believe we have just kept `Foot` because `<tfoot />` exists on HTML5. We can rename it (as well as `Head` to `Header`) in the future, but keeping `Foot` (and `Head`) as an alias with some kind of deprecation notice.

packages/fuselage-hooks/src/useLocalStorage.server.spec.ts
dougfabris: I discuss it deeper with gazzo, thanks for the explanation guys!

packages/fuselage/src/components/Table/index.d.ts
dougfabris: Makes total sense!

dougfabris (approved):

PR #573, opened 2021-11-18
fix(onboarding-ui): Remove manage workspace fallback

ujorgeleite:
remove unnecessary components on the request trial page

- [x] I have signed the CLA
- [x] I have run Loki's visual regression tests (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)


dougfabris (approved):

packages/onboarding-ui/src/common/ManageWorkspaceFallback.spec.tsx
```diff
@@ -0,0 +1,12 @@
...
+import ManageWorkspaceFallback from './ManageWorkspaceFallback';
+
+it('renders without crashing', () => {
+  const div = document.createElement('div');
+  ReactDOM.render(
+    <ManageWorkspaceFallback onManageWorkspaceClick={() => undefined} />,
+    div
+  );
+  ReactDOM.unmountComponentAtNode(div);
+});
```
geekgonecrazy, line 12: shouldn't this be removed too?

geekgonecrazy: or is the component being used again elsewhere?

ujorgeleite: We understood that maybe we use this in the future, in this way not is necessary to remove it.

PR #578, opened 2021-11-29
fix(fuselage): Option component typings

dougfabris:
- [x] I have signed the CLA
- [x] I have run Loki's visual regression tests (if applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)


tassoevan (approved):
To summarize what I've changed:

- Since it was not working as expected, changed it to a `fix`-type PR;
- I've split the component into single modules following the convention we adopt at the main repo*;
- I've fixed some references to an `Option` named export when the correct one is the default-exported.

* I think we soon should be enabling the [`react/no-multi-comp` rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md) here too.

PR #625, opened 2022-01-20
docs(fuselage): Add Field use examples

PedroRorato:
- Add more information about how to use Field components
https://www.loom.com/share/b55dd1bfec874ebfa3aa67f4a0fe8849


tassoevan (approved):
That's a good initiative and, since Loki tests are gone, we can consider doing stories as documentation only.

PR #630, opened 2022-01-27
feat(memo): Cache timeout

tassoevan:

### DIFF
--- a/packages/memo/src/memoize.ts
+++ b/packages/memo/src/memoize.ts
@@ -37,11 +37,8 @@ export const memoize = <T, A, R>(
 };
 
 export const clear = (
-  fn: MemoizableFunction<unknown, unknown, unknown>
+  fn: MemoizedFunction<unknown, unknown, unknown>
 ): void => {
-  const x = store.get(fn);
-
-  if (x) {
-    x.clear();
-  }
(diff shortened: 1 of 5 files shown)


packages/memo/src/memoize.ts
```diff
@@ -13,21 +17,46 @@ const isCachedValue = <A, R>(
...
 
   const memoized: MemoizedFunction<T, A, R> = function (this, arg) {
+    const cleanUp = (): void => {
+      cache.delete(arg);
+      cacheTimers.delete(arg);
+    };
+
     const cachedValue = cache.get(arg);
 
+    // return cachedValue if already cached
```
dougfabris, line 34: I don't think we need these comments since it's well explained by the names of the functions

dougfabris (changes requested):

packages/memo/src/memoize.ts
guijun13: Ok. Should I remove the comments in the `.spec.ts` file too?

dougfabris: Commonly the comments related to the test file is the description itself, but I don't see any problem to have comments inside the test, if you want to!

dougfabris (approved):

PR #633, opened 2022-02-01
chore(fuselage): Convert from js to ts

juliajforesti:


packages/fuselage/src/components/Menu/Menu.tsx
```diff
@@ -22,19 +51,22 @@ export const Menu = ({
 }) => {
   const mappedOptions = mapOptions(options);
   const [cursor, handleKeyDown, handleKeyUp, reset, [visible, hide, show]] =
-    useCursor(-1, mappedOptions, (args, [, hide]) => {
-      menuAction(args, options);
+    useCursor(-1, mappedOptions as OptionType[], (args, [, hide]) => {
+      console.log(args);
```
gabriellsh, line 55: oops!

packages/fuselage/src/components/Menu/Menu.tsx
```diff
@@ -22,19 +51,22 @@ export const Menu = ({
...
+      menuAction(args as OptionType, options);
```
gabriellsh, line 56: This assertion is not needed, as both types match already.

gabriellsh (changes requested):

PR #643, opened 2022-02-12
feat(fuselage): `PaginatedMultiSelect` type definitions

tiagoevanp:


MartinSchoeler (approved):

dougfabris (changes requested):
First of all the component isn't working. We had to ignore typescript because it's totally messed up. I would like to know what's the main purpose of the component functionality, cause it seems to be a copy of our `Select` component which for me it's not what we should do here. There are kind of weird things like `hugeList` that should be replaced by a more smart way to generate fake data, stories that don't make sense, and duplicated. Please, provide more info and usage examples so I can help to provide the proper solution here!

PR #694, opened 2022-04-06
feat(onboarding-ui): add function to prefill Workspace URL based on W…

PedroRorato:
- Added a function to prefill the Workspace URL field based on the Workspace Name informed

### DIFF
--- a/packages/onboarding-ui/src/forms/CreateCloudWorkspaceForm/CreateCloudWorkspaceForm.stories.tsx
+++ b/packages/onboarding-ui/src/forms/CreateCloudWorkspaceForm/CreateCloudWorkspaceForm.stories.tsx
@@ -29,20 +29,21 @@ const isValidCharacter = (domainName: string) => {
   return true;
 };
 
-const validationList = [isValidLength, isValidFirstSubstring, isValidCharacter];
-
 const isValidDomainName = async (domainName: string) => {
-  let isValid: boolean | string = true;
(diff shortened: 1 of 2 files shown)


packages/onboarding-ui/src/forms/CreateCloudWorkspaceForm/CreateCloudWorkspaceForm.stories.tsx
```diff
@@ -6,6 +6,21 @@ import WorkspaceUrlInput from './WorkspaceUrlInput';
 
 type Args = ComponentProps<typeof CreateCloudWorkspaceForm>;
 
+const isValidDomainName = async (domainName: string) => {
```
ujorgeleite, line 9: I think that makes sense to turn this validation into a list of validations, and iterate on that validating the clauses and returning true or false,  maybe creating a helper or component for todo the validations in this way.

ujorgeleite (changes requested):

packages/onboarding-ui/src/forms/CreateCloudWorkspaceForm/CreateCloudWorkspaceForm.stories.tsx
PedroRorato: Nice, how about this next commit?

ujorgeleite (approved):

PR #699, opened 2022-04-14
fix: handling visibility of options in multiselect

### LINKED ISSUE #696: Multiselect: Close button not working
**Description**
Cross button on multiselect component is not working

**Steps to [reproduce**]

1. Find the component Multiselect in storybook
2. when clicked on the cross button it does not work

**Insert an image/gif**
error

sidmohanty11:
The multiselect component had an outer `Box` which handled the visibility of the `AnimatedVisibility` when clicked. So, this was getting overlapped with the actual anchorSelect's onClick. So when we press a click outside of the anchorSelect, it responds to us by going from -> hidden -> visible in one click.

multiselect-visibility

So, with this change we handle the visibility in a state keeping track of the "click to the outer box" and changing it to hidden/visible correctly.
```javascript
if (visibility === AnimatedVisibility.VISIBLE) {
       setVisibility('hidden');
       return hide();
 }
      setVisibility('visible'); // here "visible" was coming out to be hidden, so when we click on cross we again come here
      innerRef.current?.focus();
      return show();
```

## Before

https://user-images.githubusercontent.com/73601258/163399659-e93c6c6e-8426-4f53-9c35-039ff84bbc50.mp4

## After

https://user-images.githubusercontent.com/73601258/163399695-152ad71c-30b3-4e96-8b99-c3d6baea7ee7.mp4

## Issue(s)
fixes #696

### DIFF
--- a/packages/fuselage/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/fuselage/src/components/MultiSelect/MultiSelect.tsx
@@ -117,6 +117,8 @@ export const MultiSelect = forwardRef(
     const [cursor, handleKeyDown, handleKeyUp, reset, [visible, hide, show]] =
       useCursor(index, filteredOptions, internalChanged);
 
+    const [visibility, setVisibility] = useState(visible);
+
     useEffect(reset, [filter]);
 
     const innerRef = useRef<HTMLElement>(null);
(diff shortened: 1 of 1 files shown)


dougfabris:
Thanks a lot for the contribution @sidmohanty11
I'm closing this one, because we'll follow a different approach: #759

PR #700, opened 2022-04-14
docs: Fuselage usage docs and npm links fix

### LINKED ISSUE #640: docs: how to use icons of fuselage
I was working with fuselage design system and I followed the [npm readme](https://www.npmjs.com/package/@rocket.chat/fuselage) to get me started. I jumped into using the InputBox with icons and icons didn't seem to work. I thought I did something wrong and I installed ui-kits/logos etc thinking that might be the case. When I dug into the code, I got that we need to import the icons css before using them. 

## without css import
fuselage_css

Do you think it will be helpful to state this in the npm readme itself? If so can I make a PR?

sidmohanty11:
- Added docs about "how to use fuselage icons" in a project
- Added a note section about the possible error we might face while developing an application with Next.js and Fuselage
- Fixed all the npm links on the first page of the repository

## Issue(s)
Closes #640


dougfabris (changes requested):
Let's move the additional info to `fuselage-icons` context!

sidmohanty11:
@dougfabris should I keep this part under fuselage or also shift it to icons

"
Note: If you are using Next.js for development and wish to use Rocket.Chat Fuselage components, you need to dynamically import the component. More on this can be found [here](https://dev.to/vvo/how-to-solve-window-is-not-defined-errors-in-react-and-next-js-5f97#3-third-solution-dynamic-loading).

For example,

```javascript
import dynamic from 'next/dynamic';
export const TextInput = dynamic(
  () => import('@rocket.chat/fuselage').then((module) => module.TextInput),
  { ssr: false }
);
```
"

dougfabris:
> And under icons, should I add the info like this, "
> 
> ### Usage
> To use the fuselage icons, you need to import the css first:
> 
> ```css
> @import url('@rocket.chat/icons/dist/rocketchat.css');
> ```
> 
> then you can use,
> 
> ```js
> <Icon name="hash" size="x20" />
> ```
> 
> "
> 
> Thank you

Yes, let's keep the information related to nextJS on fuselage.

About the icon, we can say we recommend the usage of the fuselage component but you can also add something like:
`<i class='rcx-icon'>{icon}</i>`

dougfabris (approved):

PR #709, opened 2022-04-28
feat(fuselage): MessageEmoji and ThreadMessageEmoji components

gabriellsh:
image
image

### DIFF
--- a/packages/fuselage/src/components/Message/MessageEmoji.tsx
+++ b/packages/fuselage/src/components/Message/MessageEmoji.tsx
@@ -1,0 +1,33 @@
+import type { ReactNode } from 'react';
+import React from 'react';
+
+import { MessageEmojiBase } from './MessageEmojiBase';
+
+type MessageEmojiProps = {
+  name: string;
+  className?: string;
+  image?: string;
+  big?: boolean;
+  children?: ReactNode;
+};
+
+export const MessageEmoji = ({
+  name,
+  className,
+  image,
+  big,
+  children,
+}: MessageEmojiProps) => (
+  <MessageEmojiBase
+    className={[
+      'rcx-message__emoji',
+      className,
+      big && 'rcx-message__emoji--big',
+    ]
+      .filter(Boolean)
+      .join(' ')}
+    name={name}
+    image={image}
+    children={children}
+  />
+);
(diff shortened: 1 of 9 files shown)


packages/fuselage/src/components/Message/MessageEmojiBase.tsx
```diff
@@ -0,0 +1,22 @@
...
+  image?: string;
+};
+
+export const MessageEmojiBase = ({
+  name,
+  className,
+  image,
+  children,
+}: MessageEmojiBaseProps) => (
+  <div
```
filipemarins, line 17:
```suggestion
  <span
```

Because it could be children of `<p>`

packages/fuselage/src/components/Message/MessageEmoji.tsx
```diff
@@ -0,0 +1,33 @@
...
+  name,
+  className,
+  image,
+  big,
+  children,
+}: MessageEmojiProps) => (
+  <MessageEmojiBase
+    className={[
+      'rcx-message__emoji',
+      className,
```
filipemarins, line 24:
```suggestion
      className && 'rcx-message__emoji',
      className,
```
Can we add a conditional here? cause when the className is undefined we will render the emoji text like `:random-text:` so we don't need the emoji style in this scenario.

filipemarins: also we need the `name` in the class name to render the correct style of `emoji`

gabriellsh: I kind of agree with the first comment. I know we need for this to be "sizeless" whenever the emoji doesn't exist, but not sure if we should depend on the className. For ex: if for some reason we decide to pass another className, this implementation falls apart. Maybe we need an aditional prop?

On the second one, I don't se a reason to add the name in the className, because that is Application specific. If the application needs to send the name as a className, it can do so by including it in the className prop.

packages/fuselage/src/components/Message/MessageEmojiBase.tsx
gabriellsh: Yeah, I changed it's display to be inline-block so this shouldn't cause problems, but I will change because it's more semantic. Thanks for the tip!

filipemarins (approved):

packages/fuselage/src/components/Message/MessageEmoji.tsx
gabriellsh: I just realized that for the first comment, we can just not use this component :p. I'll leave it like it is...

filipemarins (approved):

PR #737, opened 2022-06-03
feat(fuselage-ui-kit): add URL option to the overflow component

tapiarafael:

Allow passing an URL to the Overflow component options. When an option has the URL prop, open it in a new tab before triggering the action;

## Further comments

This feature will allow opening a link inside the overflow component, as it's possible with the button component. We will need it to develop the Google Drive app.
More info [here](https://app.clickup.com/t/2mkhme4)


tapiarafael:
Maybe it's a good time to change the ObjectType from an array to an object? I didn't do it because it'll require many changes in other components besides the Overflow element.

packages/fuselage/src/components/Options/useCursor.ts
```diff
@@ -67,7 +67,8 @@ export const useCursor = <
     value: unknown,
     label: unknown,
     selected?: unknown,
-    type?: OptionType[3]
+    type?: OptionType[3],
+    url?: unknown
```
ggazzo, line 71: why url is `unknown` ?

ggazzo (changes requested):

packages/fuselage/src/components/Options/useCursor.ts
tapiarafael: I was following the other fields and thought that since `label` was a string but it's typed as an unknown the `URL` should be equal.

ggazzo (approved):

PR #757, opened 2022-06-28
feat(fuselage): Add message status indicator text component

filipemarins:
## Further comments

Add a component to show the text status in a message.

Screen Shot 2022-06-28 at 11 43 57

This component will be used to show the translation provider on rocketchat app

Screen Shot 2022-06-28 at 12 34 49


tassoevan (changes requested):
I suggest to drop the spreading of `props` because it feels like we're falling into two code smells, one being the classic [Long Parameter List](https://refactoring.guru/pt-br/smells/long-parameter-list), and the other one is more related to TypeScript: as we are not restricting `props` shape on `MessageStatusIndicatorTextProps`, it can be anything, even properties not supported by the underlying `<div />` element.

tassoevan (approved):

PR #767, opened 2022-07-11
feat(icons): Updated contact and globe icons

filipemarins:
## Further comments

Update contact and globe icons


dougfabris (approved):

tassoevan (changes requested):
I know this is a stretch from the goal of this pull request, but I can't deal with the fact that `contact` violates our standard, stated at the README:

> The filenames define the icon names, so they should stick with a basic rule: *an icon name must describe the icon shape, not its utility*.

In this scenario, the icon should be renamed as `address-book` and an alias for it must be placed as a symlink at `src/aliases/contact.svg`.

tassoevan (approved):

PR #778, opened 2022-07-14
fix: Adjusting MessageSystem font-size

hugocostadev:


packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
```diff
@@ -45,7 +45,7 @@ $message-system-background-color-selected: functions.theme(
   }
 
   &__name {
-    @include typography.use-font-scale(h5);
+    @include typography.use-font-scale(c2);
```
gabriellsh, line 48: I think this should be c1b?

PR #781, opened 2022-07-18
fix: placeholder texts being parsed as links

aleksandernsilva:

This change adds a restriction to the parser grammar in order to prevent placeholder texts (e.g `{{visitor.name}}`) from being rendered as links.

Before:
Screen Shot 2022-07-18 at 09 38 28
After
Screen Shot 2022-07-18 at 09 39 22
:


aleksandernsilva:
We found this pull request won't be necessary anymore.

PR #800, opened 2022-08-09
fix(message-rewrite): Remove emojis inside links

hugocostadev:

https:// renders an emoji in some situations
Example: 
image

To fix the issue I changed the logic for grammar `domainName` 

Benchmark compared with `develop` branch:
image


filipemarins (changes requested):
The issue is not only about the link, the issue is the emoji is being rendered in the middle of the sentence. Examples: 
`10:30` will render  1😇0. 
`he:)llo` is rendering he😄llo.  
In both scenarios it should not render the emoji.

ggazzo:
I'm sorry but the url error is a valid one and the change actually solves the issue somehow

So I dont get the comment:

> The issue is not only about the link, the issue is the emoji is being rendered in the middle of the sentence. Examples: `10:30` will render 1😇0. `he:)llo` is rendering he😄llo. In both scenarios it should not render the emoji.

ggazzo (dismissed):

filipemarins (dismissed):

filipemarins (approved):

PR #826, opened 2022-08-16
feat: Slider component

### LINKED ISSUE #425: Slider Component
**Is the team working on any slider component?**

I was expecting a slider in the Forms part in the storybook but couldn't find. Surprisingly, project hasn't needed it till now. 
I require it for this feature:
Pausing notifications for a few minutes. 
I want the user to be able to choose using a slider, although I can use a number input as well for the time being.

**Expected behavior**
* Able to pass min and max value.
* If possible, the component should handle checkpoints/marks/step-size.

**Image**
image
image

juliajforesti:
## Issue(s)

Fixes #425

### DIFF
--- a/packages/fuselage/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/fuselage/src/components/Tabs/Tabs.stories.tsx
@@ -55,8 +55,8 @@ SelectedAndDisabled.args = {
   selected: true,
 };
 
-export const Underline: ComponentStory<typeof Tabs> = Template.bind({});
-Underline.args = {
+export const NoUnderline: ComponentStory<typeof Tabs> = Template.bind({});
+NoUnderline.args = {
   selected: true,
-  underline: true,
+  divider: false,
 };
--- a/packages/fuselage/src/components/Tabs/Tabs.styles.scss
+++ b/packages/fuselage/src/components/Tabs/Tabs.styles.scss
@@ -4,8 +4,10 @@
 
 $tabs-background-color: theme('tabs-background-color', transparent);
 $tabs-border-color: theme('tabs-border-color', transparent);
-$tabs-color: theme('tabs-color', colors.font(secondary-info));
-$tabs-hover-color: theme('tabs-hover-border-color', colors.font(titles-labels));
+$tabs-color: theme('tabs-color', colors.font(annotation));
+
+$tabs-hover-color: theme('tabs-hover-border-color', colors.font(hint));
+
 $tabs-selected-color: theme(
   'tabs-selected-color',
   colors.button(primary-default)
@@ -23,6 +25,7 @@ $tabs-hover-selected-border-color: theme(
   colors.button(primary-hover)
 );
 $tabs-active-color: theme('tabs-active-color', colors.font(titles-labels));
+
 $tabs-active-selected-color: theme(
   'tabs-active-selected-color',
   colors.button(primary-press)
@@ -97,11 +100,15 @@ $tabs-disabled-selected-border-color: theme(
   &.hover,
   &:hover {
     color: $tabs-hover-color;
+    border-block-end-color: $tabs-hover-color;
+    border-block-end-width: lengths.border-width(4);
   }
 
   &.active,
   &:active {
     color: $tabs-active-color;
+    border-block-end-color: $tabs-active-color;
+    border-block-end-width: lengths.border-width(4);
   }
 
   @include on-focus-visible {
@@ -129,14 +136,14 @@ $tabs-disabled-selected-border-color: theme(
     border-block-end-width: lengths.border-width(2);
     border-inline-width: lengths.border-width(2);
 
-    &:hover:not(.rcx-tabs__item--selected--disabled) {
+    &:hover:not(.rcx-tabs__item--selected--disabled),
+    &.hover:not(.rcx-tabs__item--selected--disabled) {
       color: $tabs-hover-selected-color;
       border-block-end-color: $tabs-hover-selected-border-color;
-      border-block-end-width: lengths.border-width(4);
     }
 
-    &.active,
-    &:active {
+    &.active:not(.rcx-tabs__item--selected--disabled),
+    &:active:not(.rcx-tabs__item--selected--disabled) {
       color: $tabs-active-selected-color;
       border-block-end-color: $tabs-active-selected-border-color;
     }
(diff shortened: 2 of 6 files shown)


packages/fuselage/src/components/Slider/Slider.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+    render(<Default />);
+
+    const slider = screen.getByRole<HTMLFormElement>('slider');
+
+    slider.focus();
+
+    fireEvent.keyDown(document.activeElement!, { key: 'ArrowRight' });
+    fireEvent.keyDown(document.activeElement!, { key: 'ArrowRight' });
+    fireEvent.keyDown(document.activeElement!, { key: 'ArrowRight' });
+    fireEvent.keyDown(document.activeElement!, { key: 'ArrowRight' });
```
tassoevan, line 45: **nitpick:** Maybe there is some value in replacing `fireEvent` with `userEvent`, as it emulates human interaction in a clearer API.

packages/fuselage/src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+    <Slider
+      multiThumb
+      label='Price range'
+      defaultValue={[100, 350]}
+      maxValue={500}
+      step={10}
+      formatOptions={{ style: 'currency', currency: 'USD' }}
+    />
+  </Box>
+);
```
tassoevan, line 96: **nitpick:** What about an example story with a play function?

packages/fuselage/src/components/Slider/Slider.styles.scss
```diff
@@ -0,0 +1,96 @@
+@use '../../styles/colors.scss';
```
tassoevan, line 1: **question:** Are we supposed to keep using Sass?

packages/fuselage/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
+import type { AriaAttributes } from 'react';
+import React, { useRef } from 'react';
+import { useNumberFormatter, useSlider } from 'react-aria';
+import { useSliderState } from 'react-stately';
+
+import { SliderHead } from './SliderHead';
+import { SliderThumb } from './SliderThumb';
+import { SliderTrack } from './SliderTrack';
+
+type SliderProps<T = number | number[]> = AriaAttributes & {
```
tassoevan, line 10: If `multiThumb` is about having *two* thumbs, we should use a stricter constraint for `T`:

```suggestion
type SliderProps<T = number | [min: number, max: number]> = AriaAttributes & {
```

packages/fuselage/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+   * @default false
+   */
+  multiThumb?: T extends number[] ? true : false;
+  step?: number;
+  /**
+   * @default 0
+   */
+  minValue?: number;
+  /**
+   * @default 100
```
tassoevan, line 28: **praise:** very nice usage of JSDoc here!

packages/fuselage/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+   */
+  maxValue?: number;
+  orientation?: 'horizontal' | 'vertical';
+  isDisabled?: boolean;
```
tassoevan, line 32: `is-` prefix is fine for method names, maybe for variable names also, but it becomes odd for object properties.

```suggestion
  disabled?: boolean;
```

packages/fuselage/src/components/Slider/SliderHead.tsx
```diff
@@ -0,0 +1,48 @@
...
+          {state.getThumbValueLabel(0)}
+        </Label>
+      )}
+    </Label>
+  );
+  return (
+    <>
+      {label ? (
+        <div className='rcx-slider-label-container'>
+          <Label {...(labelProps as any)}>{label}</Label>
```
tassoevan, line 40: **question:** What is up with all these `as any` spreads?

packages/fuselage/src/components/Slider/Slider.styles.scss
```diff
@@ -0,0 +1,96 @@
...
+  height: 100%;
+
+  transform: translateX(-50%);
+}
+
+.rcx-slider-thumb {
+  width: 16px;
+  height: 16px;
+
+  cursor: pointer;
```
tassoevan, line 72: I'd like it to be `default` when the slider is disabled.

tassoevan (changes requested):

packages/fuselage/src/components/Slider/Slider.tsx
juliajforesti: that's the prop name expected by `useSlider`... 
I could manipulate it before passing if you think it's necessary

packages/fuselage/src/components/Slider/Slider.tsx
juliajforesti: totally agree. But this type is also what the lib expects and I had trouble when I typed it like you suggested. 
could you help with this?

packages/fuselage/src/components/Slider/SliderHead.tsx
juliajforesti: `Label` was giving me a ts error because of the `is` prop spread on outputProps
_ Types of property 'is' are incompatible._

packages/fuselage/src/components/Slider/Slider.spec.tsx
juliajforesti: didn't find how to fire a userEvent for "dragging" that worked.

packages/fuselage/src/components/Slider/SliderHead.tsx
juliajforesti: Done: Omitted 'is' prop

packages/fuselage/src/components/Slider/SliderThumb.tsx
```diff
@@ -0,0 +1,53 @@
+import { css } from '@rocket.chat/css-in-js';
+import '../../styles/colors.scss';
```
tassoevan, line 2:
```suggestion
```

tassoevan (changes requested):

tassoevan (approved):

PR #828, opened 2022-08-17
test(fuselage): Add test cases to Callout component

PedroRorato:
- Add helper function to simplify setting Stories description;
- Add test cases to Callout component


packages/fuselage/src/components/Callout/Callout.spec.tsx
```diff
@@ -1,13 +1,52 @@
...
 import * as stories from './Callout.stories';
 
-const { Default } = composeStories(stories);
+const { Default, WithDescriptionOnly, Info, Success, Warning, Danger } =
+  composeStories(stories);
 
 describe('[Callout Component]', () => {
-  it('renders without crashing', () => {
-    render(<Default />);
+  describe('Story renders without crashing', () => {
```
dougfabris, line 12: I think just the `Default` variant is enough for this first test

packages/fuselage/src/components/Callout/Callout.stories.tsx
```diff
@@ -36,36 +37,53 @@ export default {
...
-      {args.children}
-    </Callout>
-    <Callout title='This is a warning message' type='warning'>
-      {args.children}
-    </Callout>
-    <Callout title='This is a danger message' type='danger'>
-      {args.children}
-    </Callout>
-  </>
+  <Callout {...args} />
```
dougfabris, line 40: Since all stories are going to have the same `children`, I would suggest adding it to the template.

dougfabris (changes requested):

packages/fuselage/src/components/Callout/Callout.stories.tsx
PedroRorato: You are right, I agree.

packages/fuselage/src/components/Callout/Callout.spec.tsx
PedroRorato: I did some refactoring

PR #830, opened 2022-08-19
fix: Remove italic style from system messages

hugocostadev:

The design team raised an issue to remove the italic style from the system message component. 
It was removed from the `<MessageSystem>` component but there are several messages with other font styles (italic and bold) at [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) messages that are being checked with the Design team if they should be removed, but it not an blocker.
 

## Further comments

TO DO: 

- [ ] If the other font-styles should be removed from [Rocket.Chat](https://github.com/RocketChat/Rocket.Chat) the PR will be included at main repository


ggazzo (changes requested):
could you provide me more info about this decision?

hugocostadev:
> could you provide me more info about this decision?

Already aligned in private

gabriellsh (approved):

ggazzo (approved):

PR #839, opened 2022-08-30
feat: Adding new Styles for Quote Attachments

hugocostadev:

Updating QuoteAttachment styles and structure, now it's using the same wrapper (`<MessageBody />`) as the normal messages

This Pull Request only adds some styles to the existing `<QuoteAttachment />` at Rocket.Chat core repository

The PR that changes the structures is: https://github.com/RocketChat/Rocket.Chat/  

**IMPORTANT**: For the feature to be considered DONE both PRs need to be merged. I suggest doing tests and QA on the [local environment syncing both branches](https://github.com/RocketChat/fuselage/wiki/Testing-fuselage's-components-on-Rocket.Chat) 

**Actual behavior**
image

**Expected behavior**
image

### DIFF
--- a/packages/fuselage/src/components/Message/mixins.scss
+++ b/packages/fuselage/src/components/Message/mixins.scss
@@ -1,4 +1,5 @@
 @use '../../styles/lengths.scss';
+@use '../../styles/colors.scss';
 
 @mixin container() {
   &-container {
@@ -25,6 +26,14 @@
 
       width: lengths.size(36);
       margin-block: lengths.margin(-2);
+    }
+  }
+
+  & .rcx-attachment {
+    &__details {
+      .rcx-message-body {
+        color: colors.neutral(700);
+      }
     }
   }
 }


packages/fuselage/src/components/Message/mixins.scss
```diff
@@ -27,6 +28,14 @@
       margin-block: lengths.margin(-2);
     }
   }
+
+  & .rcx-attachment {
+    &__details {
```
filipemarins, line 33: The `.rcx-attachment__details` is an app component. I don't think we should have app style here on fuselage. I guess the right place for this [style is here](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/components/message/Attachments/QuoteAttachment.tsx#L17).

packages/fuselage/src/components/Message/mixins.scss
```diff
@@ -27,6 +28,14 @@
...
+      .rcx-message-body {
+        color: colors.neutral(700);
```
filipemarins, line 35: The `.rcx-attachment__details` is an app component. I don't think we should have app style here on fuselage. I guess the right place for this [style is here](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/client/components/message/Attachments/QuoteAttachment.tsx#L17).

hugocostadev: Yep, you're right, thanks @filipemarins .... closing this PR to include this change at [Core Pull Request](https://github.com/RocketChat/Rocket.Chat/pull/26751)

PR #856, opened 2022-09-29
feat(fuselage): `Modal` wrapper option

dougfabris:


tassoevan (changes requested):
I guess you tried the approach of passing an element as a prop and cloning it before in another situation, when there was some pitfalls. I would prefer `wrapper` to be a render function, and maybe have an additional `wrapperComponent` prop.

tassoevan (approved):

PR #859, opened 2022-09-30
refactor: remove unimplemented components from index

tapiarafael:

Remove unimplemented components from the index file, so they don't appear on the intellisense.

## Further comments
This change is made to avoid confusion during the development using ui-kit, since the components would not work as they aren't implemented yet.

[Task AE-56](https://rocketchat.atlassian.net/browse/AE-56)


ggazzo:
why not implement the components? since now we moved the fuselage-ui-kit to the main repo, thing are much easier to be done..

tassoevan:
I believe it's fine. The exposed API should really reflect what we have at the moment. Future components are better described in spec files.

PR #865, opened 2022-10-11
fix(fuselage): Tag not receiving props properly

dougfabris:


gabriellsh (changes requested):
The correct fix for this case would be to fix the type annotation. We shouldn't use `Box` in this case, because at one point this component was rewritten to remove Box from it's dependencies (probably due to performance).

gabriellsh (approved):

PR #889, opened 2022-11-01
feat(fuselage): add message names wrapper and update system message

filipemarins:

Add `MessageNames` wrapper to display `name` and `username` inline and separate them with a blank space.

Update message system font size:

Before:
Screenshot 2022-11-01 at 16 12 38

After:
Screenshot 2022-11-01 at 16 16 28

Add system variation to ThreadOrigin component
Screenshot 2022-10-31 at 14 05 09

## Further comments

The system variation on ThreadOrigin component will be used to show the message removed status.


gabriellsh (changes requested):
Looks good! Only suggestion is to change `MessageNames` to something more descriptive, since it can be confused with `MessageName`. Perhaps `MessageNameContainer`?

gabriellsh (dismissed):

gabriellsh (approved):

PR #891, opened 2022-11-04
fix(fuselage): Using Pre Wrap For System Message

nishant23122000:
### Before :
Screenshot 2022-11-11 at 12 33 13 AM

### After : 
Screenshot 2022-11-11 at 12 33 55 AM

reason of change :  https://github.com/RocketChat/Rocket.Chat/pull/27080

### DIFF
--- a/packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
+++ b/packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
@@ -38,7 +38,7 @@ $message-system-background-color-selected: functions.theme(
     @include typography.use-font-scale(p2);
     @include typography.use-with-truncated-text();
     margin-inline: lengths.margin(2);
-
+    white-space: pre-wrap;
     color: colors.font(annotation);
   }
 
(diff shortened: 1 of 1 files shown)


nishant23122000:
> Can you add screenshots so we now how it changed visually? Also a description so we could understand the issue better would help. Thanks!

Done check now

gabriellsh:
Hey @nishant23122000 ! I'll be checking with design to see if this is the desired UI/UX for this. If we're all in agreement, I'll review the code and merge this.

jsdgdo:
Hello @nishant23122000, José from Rocket.Chat's design team here. I was wondering **what prompted your need to work on longer system messages**, in our experience they usually are no longer than 4 or 5 words. So we don't actually see a need for this particular improvement. 

Either way I'm excited to find a solution to this problem in the coming releases, but this particular one doesn't agree with the spacing standards we set up for the platform. Hope you understand.

nishant23122000:
If words are less than 4 or 5 then it makes sense to don't do these changes.

but I think the agent can write long sentences also in some cases.

dudanogueira:
The UX problem, IMHO, is that the receiving agent will not have access to that comment in full, which is usually very important for the chat handling.

Isn't there a way to allow that message to be shown, like on mouse hover?

PR #910, opened 2022-11-22
fix(fuselage): Chevron vertical position

vatsilidis:
Add vertical center alignment at the Chevron in order to avoid positioning it at the top or at the bottom when the height of the parent container is bigger than the Chevron height.

Here are two screenshots with a title of height of 60px at the Accordion's header before and after the fix.

Screenshot 2022-11-22 at 08 52 23
Screenshot 2022-11-22 at 08 52 15

Screenshot 2022-11-22 at 08 51 26
Screenshot 2022-11-22 at 08 51 22


ggazzo:
help @tassoevan 
not sure if makes sense or not, remember this component is not only used with subtitle
on rocket.chat there are some with complex (and huge forms) having the chevron centralized could be a little weird in those cases

tassoevan (approved):

tassoevan:
It makes sense to reinforce the vertical centralization, but we *don't* recommend using anything beyond a plain text as the accordion's title.

PR #922, opened 2022-11-30
feat(fuselage): Featured Tag variant

dougfabris:
Screen Shot 2022-11-30 at 19 20 59

It also fixes the ellipsis issue
Screen Shot 2022-11-30 at 19 22 28


packages/fuselage/src/components/Tag/Tag.styles.scss
```diff
@@ -94,10 +107,7 @@ $tag-colors-disabled-background-color: theme(
 );
 
 .rcx-tag {
-  display: flex;
```
juliajforesti, line 97: Removing this, misaligns the Icon and the text

juliajforesti (changes requested):

juliajforesti (approved):

PR #937, opened 2022-12-12
feat(onboarding-ui): Loading state on workspace first member creation form

csuadev:

Add a loading state to the workspace first member creation form to give feedback to the user when the request takes some time. 

## Issue(s)

[https://rocketchat.atlassian.net/browse/CS-272](https://rocketchat.atlassian.net/browse/CS-272)


geekgonecrazy (approved):

geekgonecrazy:
Approved from general frontend perspective.  This code is being used by cloud portal and we need to be able to show that some action is happening.  Without it we are seeing people drop because they think the UI got stuck

PR #939, opened 2022-12-15
feat(fuselage): add disabled variant to option

filipemarins:
## Further comments

Add disabled variant to `Option` component, also move `prevent` to a helper function since it was repeating too much.

### DIFF
--- a/packages/fuselage/src/components/Option/Option.stories.tsx
+++ b/packages/fuselage/src/components/Option/Option.stories.tsx
@@ -168,7 +168,7 @@ export const WithIcon: ComponentStory<typeof Option> = () => (
   </Box>
 );
 
-export const WithDisabled: ComponentStory<typeof Option> = () => (
+export const Disabled: ComponentStory<typeof Option> = () => (
   <Box position='relative' maxWidth={330}>
     <Option onClick={action('click')}>
       <OptionContent>Enabled</OptionContent>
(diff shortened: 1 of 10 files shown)


aleksandernsilva (dismissed):

packages/fuselage/src/components/Option/Option.styles.scss
```diff
@@ -109,6 +109,10 @@ $variants: (
     background: colors.surface(selected);
   }
 
+  &--disabled {
+    color: colors.font(disabled);
```
dougfabris, line 113:
```suggestion
    cursor: not-allowed;
    
    color: colors.font(disabled);
```

packages/fuselage/src/components/Option/Option.stories.tsx
```diff
@@ -168,6 +168,17 @@ export const WithIcon: ComponentStory<typeof Option> = () => (
   </Box>
 );
 
+export const WithDisabled: ComponentStory<typeof Option> = () => (
```
dougfabris, line 171:
```suggestion
export const Disabled: ComponentStory<typeof Option> = () => (
```

dougfabris (changes requested):

packages/fuselage/src/components/Option/Option.styles.scss
filipemarins: nice

dougfabris (approved):

aleksandernsilva (approved):

PR #942, opened 2022-12-27
feat(fuselage): add aria-disabled prop to Option component

filipemarins:
## Further comments

Add `aria-disabled` prop to Option component

### DIFF
--- a/packages/fuselage/src/components/Option/Option.tsx
+++ b/packages/fuselage/src/components/Option/Option.tsx
@@ -49,6 +49,7 @@ const Option = memo(
       id={id}
       ref={ref}
       aria-selected={selected}
+      aria-disabled={disabled}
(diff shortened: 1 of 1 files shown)


packages/fuselage/src/components/Option/Option.tsx
```diff
@@ -49,6 +49,7 @@ const Option = memo(
       id={id}
       ref={ref}
       aria-selected={selected}
+      aria-disabled={disabled}
```
tassoevan, line 52: I guess it should be an string ("true" or "false"):

```suggestion
      aria-disabled={String(disabled)}
```

tassoevan (changes requested):

tassoevan (approved):

PR #957, opened 2023-01-23
fix(fuselage-hooks): Overlays/popups using usePosition hook shaking

Educg550:

A fix was made on `usePosition` hook, which had a loop behavior when any frame that used it on the main application would get off the screen.

image

## Issue(s)

Message ellipsis dropdown window shakes depending on the size of the screen. They should not shake, which was resolved


gabriellsh (approved):
I'll approve this with a comment:

This changes the behaviour of the hook as it was before this changes. Before, popover's would follow the reference component when scrolling. With these changes, this no longer happens.

This could also impact the behaviour when the target (popover) changes size.

As a last resort we can revert this back if we find out this affects other behaviours.

PR #967, opened 2023-01-26
fix(fuselage-hooks): Fix `usePosition`

ggazzo:
Closes #957

### DIFF
--- a/packages/fuselage-hooks/src/usePosition.ts
+++ b/packages/fuselage-hooks/src/usePosition.ts
@@ -278,7 +278,7 @@ export const getVariantBoundaries = ({
  * @public
  */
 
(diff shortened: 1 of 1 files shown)


packages/fuselage-hooks/src/usePosition.ts
```diff
@@ -298,7 +278,7 @@ export const getVariantBoundaries = ({
  * @public
  */
 
-export const usePosition = <T extends Element, R extends Element>(
+export const usePosition = <T extends HTMLElement, R extends HTMLElement>(
```
dougfabris, line 281:
```suggestion
export const usePosition = <T extends Element, R extends Element>(
```

dougfabris (changes requested):

dougfabris (approved):

PR #979, opened 2023-02-14
refactor(fuselage): Move margin inline style to inline box at `Modal`

filipemarins:
## Further comments

Right now we don't have fallback styles for `margin-inline` compatible with older browsers.

styles in safari 14:
Screenshot 2023-02-14 at 11 27 38

### DIFF
--- a/packages/fuselage/src/components/Modal/ModalContent.tsx
+++ b/packages/fuselage/src/components/Modal/ModalContent.tsx
@@ -15,7 +15,7 @@ export const ModalContent = ({
 }: ModalContentProps) => (
   <Scrollable vertical onScrollContent={onScrollContent}>
     <Box rcx-modal__content>
-      <Box rcx-modal__content-wrapper {...props}>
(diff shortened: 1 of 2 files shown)


packages/fuselage/src/components/Modal/ModalContent.tsx
```diff
@@ -15,7 +15,7 @@ export const ModalContent = ({
 }: ModalContentProps) => (
   <Scrollable vertical onScrollContent={onScrollContent}>
     <Box rcx-modal__content>
-      <Box rcx-modal__content-wrapper {...props}>
+      <Box rcx-modal__content-wrapper marginInline='x24' {...props}>
```
tassoevan, line 18:
```suggestion
      <Box rcx-modal__content-wrapper mi={24} {...props}>
```

tassoevan (changes requested):

tassoevan (approved):

PR #990, opened 2023-02-23
chore(fuselage): fix ModalIcon vertical align

juliajforesti:
[ARCH-149](https://rocketchat.atlassian.net/browse/ARCH-149)

## Further comments

[ARCH-149]: https://rocketchat.atlassian.net/browse/ARCH-149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/fuselage/src/components/Modal/Modal.styles.scss
+++ b/packages/fuselage/src/components/Modal/Modal.styles.scss
@@ -48,6 +48,7 @@ $modal-margin: theme('modal-margin', auto);
   &__header-inner {
     display: flex;
     flex-wrap: nowrap;
+    align-items: center;
 
     margin: -4px;
   }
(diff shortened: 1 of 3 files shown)


packages/fuselage/src/components/Modal/Modal.styles.scss
```diff
@@ -48,6 +48,7 @@ $modal-margin: theme('modal-margin', auto);
   &__header-inner {
     display: flex;
     flex-wrap: nowrap;
+    align-items: center;
```
dougfabris, line 51: Unfortunately, if we use this approach it's going to break the Modal with `Tagline` and `Icon`

dougfabris (changes requested):

PR #1013, opened 2023-03-21
refactor: grammar optmization

hugocostadev:

**Changes:**
- Updated the `peggy` library to `3.0.2` version
- Simplified some operations 
- Reorder and added some comments to the code for better readability 

The performance increase is **63.6%**
**Develop branch Benchmark:**
image

**Refactor Branch Benchmark:**
image

## Issue(s)

fix https://github.com/RocketChat/Rocket.Chat/issues/28251


packages/message-parser/src/grammar.pegjs
```diff
@@ -77,206 +82,154 @@ Code_line
...
+ *
+ */
+References
+  = "[" title:LinkTitle* "](" href:LinkRef ")" { return title.length ? link(href, reducePlainTexts(title)) : link(href); }
+  / "<" href:LinkRef "|" title:LinkTitle2 ">" { return link(href, [plain(title)]); }
 
-Image
-  = "![](" href:LinkRef ")" { return image(href); }
-  / "![" title:LinkTitle "](" href:LinkRef ")" { return image(href, title); }
+LinkTitle = @(Emphasis / Line / Whitespace)
```
polkovnikov-ph, line 232: It's the only item in sequence, not even `@` is needed

packages/message-parser/src/grammar.pegjs
```diff
@@ -287,101 +240,55 @@ LinkTitle2
...
-  / ":"
-  / ";"
-  / "("
-  / ")"
-  / "="
-  / "~"
+URL_authorityPassword = $(alpha_digit / ![@/] safe)+
 
-hexdigit = [0-9A-Fa-f]
+URL_authorityHost = t:URL_authorityHostName (":" URL_authorityPort)?
```
polkovnikov-ph, line 269: `t:` is extraneous

packages/message-parser/src/grammar.pegjs
```diff
@@ -667,27 +543,84 @@ unicodeEmojiDingbats = $([\u2700-\u27BF] [\uFE00-\uFE0F]?)
...
+
+hexByte = a:hexdigit b:hexdigit { return parseInt(a + b, 16); }
+
+alphaNumericOrMarkChar = alphaOrMarkChar / decimalNumberChar
 
-/* Tasks */
+alphaOrMarkChar = alphaChar / emojiChar / markChar
 
-tasks = items:task+ { return tasks(items); }
+decimalNumberChar = [0-9\u0660-\u0669\u06F0-\u06F9\u07C0-\u07C9\u0966-\u096F\u09E6-\u09EF\u0A66-\u0A6F\u0AE6-\u0AEF\u0B66-\u0B6F\u0BE6-\u0BEF\u0C66-\u0C6F\u0CE6-\u0CEF\u0D66-\u0D6F\u0DE6-\u0DEF\u0E50-\u0E59\u0ED0-\u0ED9\u0F20-\u0F29\u1040-\u1049\u1090-\u1099\u17E0-\u17E9\u1810-\u1819\u1946-\u194F\u19D0-\u19D9\u1A80-\u1A89\u1A90-\u1A99\u1B50-\u1B59\u1BB0-\u1BB9\u1C40-\u1C49\u1C50-\u1C59\uA620-\uA629\uA8D0-\uA8D9\uA900-\uA909\uA9D0-\uA9D9\uA9F0-\uA9F9\uAA50-\uAA59\uABF0-\uABF9\uFF10-\uFF19]
```
polkovnikov-ph, line 620: I suspect these merged character classes are the main reason why performance got so much better

packages/message-parser/src/grammar.pegjs
hugocostadev: Why not? The `LinkTitle` should get the  returned value, to make it work there are two options:

`LinkTitle = text:(Emphasis / Line / Whitespace) { return text; }`
OR as it is:
`LinkTitle = @(Emphasis / Line / Whitespace)`

Just the syntax is different, right? Is there any major differences/trade-offs/performance impact between them?

packages/message-parser/src/grammar.pegjs
hugocostadev: The `decimalNumberChar` or the `tasks`? 
The `decimalNumberChar` was already like that... 

Maybe the perf increase was caused by some use of `count`, `min` and `max` rules, rewrites of some logic, simplifications of `AND` characters operations using regex (e.g:  `safe`, `extra` and etc...)

I'm still trying to find ways to improve and I'm investigating a better way to build emojis/emoticons as you suggested!

BTW, @polkovnikov-ph  thanks a lot for your contribution and your interest

polkovnikov-ph: No problem at all. It didn't take much time, and I'm a huge fan of user interfaces that are fast and responsive :)

guijun13 (approved):

PR #1065, opened 2023-05-31
feat(UiKit): Add `I18N` structure to `TextObject`

tiagoevanp:
## Further comments

AECO-195


tassoevan:
I don't like the idea of introducing a new text object for that... Is it going to support gazzodown? Are there any other rendering dissimilarities from `plain_text` and `mrkdwn`?

tassoevan (approved):

PR #1072, opened 2023-06-06
chore(fuselage): add `Option` gap

juliajforesti:

When using `Option` you can use the `gap` prop to add spacing to the left.
If the list is mixed (items **with and without** icons, you should add the gap to the elements without icons.

image


dougfabris (approved):

dougfabris (changes requested):
Let's add the `<Tile />` as a `StoryDecorator` ?

dougfabris (approved):

PR #1109, opened 2023-07-17
chore(fuselage): `MenuV2` custom button

juliajforesti:


packages/fuselage/src/components/Menu/V2/Menu.tsx
```diff
@@ -28,6 +28,7 @@ interface MenuButtonProps<T> extends AriaMenuProps<T>, MenuTriggerProps {
   className?: ComponentProps<typeof Box>['className'];
   pressed?: boolean;
   maxWidth?: string;
+  customButton?: React.ReactElement;
```
tassoevan, line 31: `nitpick` Could it be just `button`?

tassoevan (changes requested):

packages/fuselage/src/components/Menu/V2/Menu.tsx
juliajforesti: I was wondering between these options:
1. menuButton
2. button
3. customButton
4. other?

imo `customButton` describes it more accurately, but i'm ok with `button` too. 
wdyt?

tassoevan: Which would be the most familiar one for someone that knows Fuselage?

juliajforesti: hummm, following the same reasoning as the `icon` being the option to chose a different one from the default (kebab), I agree with `button` :)

tassoevan (approved):

PR #1113, opened 2023-07-20
chore(fuselage): Drop dot notation for `Tabs` component

tiagoevanp:

### DIFF
--- a/packages/fuselage/src/components/Tabs/Tabs.tsx
+++ b/packages/fuselage/src/components/Tabs/Tabs.tsx
@@ -2,6 +2,7 @@ import type { ComponentProps } from 'react';
 import React from 'react';
 
 import Box from '../Box';
+import { TabsItem } from './TabsItem';
 
 type TabsProps = ComponentProps<typeof Box> & { divider?: boolean };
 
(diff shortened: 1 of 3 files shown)


packages/fuselage/src/components/Tabs/Tabs.tsx
```diff
@@ -15,5 +14,3 @@ export function Tabs({ children, divider = true, ...props }: TabsProps) {
     </Box>
   );
 }
-
-Tabs.Item = TabsItem;
```
juliajforesti, line 19: right now this will break all its usage. 
i suggest keeping it there and adding the deprecated sign

juliajforesti (changes requested):

packages/fuselage/src/components/Tabs/Tabs.tsx
```diff
@@ -16,4 +16,5 @@ export function Tabs({ children, divider = true, ...props }: TabsProps) {
   );
 }
 
+/** @deprecated dropping dot notation */
```
juliajforesti, line 19:
```suggestion
/** @deprecated use named export TabsItem instead */
```

juliajforesti (changes requested):

juliajforesti (approved):

PR #1124, opened 2023-08-05
chore: refactor length tokens

ggazzo:


dougfabris (approved):

packages/fuselage/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "rootDirs": ["./src/", "./.storybook/"],
     "target": "ES5",
     "module": "ESNext",
-    "lib": ["ES2020", "DOM"],
+    "lib": ["ES2020"],
```
tassoevan, line 7: Why?

tassoevan (changes requested):

tassoevan:
Should we enforce a "spacing" type for margins and paddings? This is quite common in utility-focused libraries like Tailwind and Styled System.

packages/fuselage/tsconfig.json
dougfabris: fixed!

tassoevan (approved):

PR #1144, opened 2023-08-24
chore(fuselage): `Button` size

juliajforesti:

- change `Button` default size to 40px

### DIFF
--- a/packages/fuselage/src/components/Button/Button.spec.tsx
+++ b/packages/fuselage/src/components/Button/Button.spec.tsx
@@ -1,10 +1,15 @@
 import { composeStories } from '@storybook/testing-react';
 import { render } from '@testing-library/react';
+import { axe, toHaveNoViolations } from 'jest-axe';
 import React from 'react';
 
 import * as stories from './Button.stories';
+import * as iconButtonStories from './IconButton.stories';
 
 const { Default, AsIconButton } = composeStories(stories);
+const { _IconButton, _IconButtonInfo, _IconButtonSuccess } =
+  composeStories(iconButtonStories);
+expect.extend(toHaveNoViolations);
 
 describe('[Button Component]', () => {
   it('renders Button without crashing', () => {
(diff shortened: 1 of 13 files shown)


packages/fuselage/src/components/ButtonGroup/ButtonGroup.spec.tsx
```diff
@@ -1,13 +1,22 @@
 import { composeStories } from '@storybook/testing-react';
 import { render } from '@testing-library/react';
+import { axe, toHaveNoViolations } from 'jest-axe';
 import React from 'react';
 
 import * as stories from './ButtonGroup.stories';
 
 const { Default } = composeStories(stories);
```
tassoevan, line 8: Jest extensions should be in a setup file, no?

packages/fuselage/src/components/Button/Button.tsx
```diff
@@ -41,7 +41,7 @@ export const Button = forwardRef(function Button(
     children,
     ...props
   }: ButtonProps,
-  ref: Ref<HTMLElement>
+  ref: Ref<HTMLButtonElement>
```
tassoevan, line 44: What if `<Box is='a' ref={ref} />`?

packages/fuselage/src/components/Button/Button.spec.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { axe, toHaveNoViolations } from 'jest-axe';
 import React from 'react';
 
 import * as stories from './Button.stories';
+import * as iconButtonStories from './IconButton.stories';
 
 const { Default, AsIconButton } = composeStories(stories);
+const { _IconButton, _IconButtonInfo, _IconButtonSuccess } =
+  composeStories(iconButtonStories);
+expect.extend(toHaveNoViolations);
```
tassoevan, line 12: Jest extensions should be in a setup file, no?

tassoevan (changes requested):

packages/fuselage/src/components/Button/Button.tsx
juliajforesti: you're right ✅ 
added ` | HTMLAnchorElement` type to ref and `AllHTMLAttributes<HTMLButtonElement | HTMLAnchorElement>`

PR #1159, opened 2023-09-11
fix(fuselage): `AudioPlayer`: `Infinity` duration audio files crashing

yash-rajpal:
## Further comments

The issue is that some kind of audio files are not returning the correct duration with metaData, they are returning `Infinity` which is crashing the UI.

So this is a common issue and we have a few workarounds for solving this - 

1. Fetch the entire file in background and then get the duration. Didn't use this approach as it would unnecessarily download(fetch) the entire file.

2. Seek the audio element to play some audio by setting the `currentTime` of the element, set this `currentTime` to some very big number, which will cause it to fetch the real duration of the audio file and then we can use this duration. I've used this approach, but I ran into another issue when using this approach. After setting the `currentTime` to a very big number, the audio element was breaking for no apparent reason. It was not seeking the audio current time to any desired location, so had to create a new Audio element and use this new element to fetch the duration.

### DIFF
--- a/packages/fuselage/src/components/AudioPlayer/AudioPlayer.tsx
+++ b/packages/fuselage/src/components/AudioPlayer/AudioPlayer.tsx
@@ -28,6 +28,32 @@ function forceDownload(url: string, fileName?: string) {
   xhr.send();
 }
 
+const getDurationForInfinityDurationAudioFile = (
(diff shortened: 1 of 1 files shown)


dougfabris (approved):
LGTM!

PR #1190, opened 2023-10-09
feat(fuselage): Added property `addon` to SidebarBanner and updated styles

aleksandernsilva:

- Changed text color from grey to white.
- Added a new property `addon`. 
Just like the property of same name available on `TextBox` component, this property allows you to add a react element to a specific slot in the component.

Before:
Screen Shot 2023-10-09 at 11 03 23

After:
Screen Shot 2023-10-09 at 11 01 39

With addon:
Screen Shot 2023-10-09 at 11 02 31

With addon and onClose:
Screen Shot 2023-10-13 at 12 36 46

### DIFF
--- a/packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
+++ b/packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
@@ -99,7 +99,7 @@ $sidebar-banner-background-default: theme(
 );
 $sidebar-banner-color-default: theme(
   'sidebar-banner-color-default',
-  colors.font(default)
+  colors.font(pure-white)
 );
 
 $sidebar-banner-background-info: theme(
@@ -367,6 +367,7 @@ $sidebar-banner-color-danger: theme(
 
     color: $sidebar-banner-color-default;
     background-color: $sidebar-banner-background-default;
+    column-gap: 0.25rem;
 
     &--text {
       @include typography.use-font-scale(h5);
@@ -374,10 +375,13 @@ $sidebar-banner-color-danger: theme(
 
     &--description {
       @include typography.use-font-scale(p2m);
+      display: inline-block;
+
       color: currentColor;
 
       &--clickable {
         cursor: pointer;
+        border-block-end: 1px solid;
       }
     }
 
--- a/packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
+++ b/packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
@@ -99,7 +99,7 @@ $sidebar-banner-background-default: theme(
 );
 $sidebar-banner-color-default: theme(
   'sidebar-banner-color-default',
-  colors.font(pure-white)
+  colors.font(titles-labels)
 );
 
 $sidebar-banner-background-info: theme(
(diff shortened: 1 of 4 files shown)


packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
```diff
@@ -367,17 +367,21 @@ $sidebar-banner-color-danger: theme(
...
 
     &--description {
       @include typography.use-font-scale(p2m);
+      display: inline-block;
+
       color: currentColor;
 
       &--clickable {
         cursor: pointer;
+        border-block-end: 1px solid;
```
ggazzo, line 384: use `lengths.border-width(default)`

packages/fuselage/src/components/Sidebar/Sidebar.styles.scss
```diff
@@ -99,7 +99,7 @@ $sidebar-banner-background-default: theme(
 );
 $sidebar-banner-color-default: theme(
   'sidebar-banner-color-default',
-  colors.font(default)
+  colors.font(pure-white)
```
ggazzo, line 102: O dont think this is the right way ,  please @juliajforesti @dougfabris  could you check plz?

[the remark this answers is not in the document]
aleksandernsilva: @juliajforesti What would be the appropriate way of changing the font color?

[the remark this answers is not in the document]
juliajforesti: what do you mean by appropriate way?
the "way" is correct, but the token `pure-white` (meaning **always white**, on dark and light theme) cannot be applied on this surface (`surface-neutral`), because on light mode it has no contrast:
screen

was this change requested by the design team? if so, I suggest bringing this issue back to discussion with them, reminding that we have dark and light theme to be checked when we change colors :)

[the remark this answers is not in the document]
juliajforesti: @aleksandernsilva font `titles-labels` is darker on light-mode and "whiter", maybe that serves the purpose?

packages/fuselage/src/components/Sidebar/SidebarBanner.tsx
```diff
@@ -42,6 +44,7 @@ export const SidebarBanner = ({
       )}
       {children}
     </div>
+    {addon}
```
juliajforesti, line 47: what happens if you add an `addon` **and** `onClose` action?
is that situation wanted or should we prevent it?

juliajforesti (commented):
I also suggest adding this case to storybook :) 
there's a Banner Variations section

packages/fuselage/src/components/Sidebar/SidebarBanner.tsx
aleksandernsilva: My intention was to not affect existing functionality. Here's an example with the `onClose`:
Screen Shot 2023-10-13 at 12 36 46

juliajforesti (approved):

PR #1201, opened 2023-10-19
fix(fuselage): `withTitle` prop not working on MultiSelect and PaginatedMultiSelect

aleksandernsilva:


dougfabris (changes requested):
This addition doesn't seem valuable to me, you can achieve this using the `renderItem` prop.
The title doesn't make sense, since we do not have `withTitle` prop there's no way it to not be working
Also, I would not recommend adding a tooltip to the `SelectOption`

PR #1206, opened 2023-10-24
fix(fuselage): change `Callout` text color

juliajforesti:

- change `Callout` text color to font-default


dougfabris (approved):
Thanks!

Issue #1298, opened 2024-02-24
AutoComplete, ToggleSwitch : bugs, feature add


umangutkarsh:
**Description**
Below are the issues and a basic transition add requested in the following components:

- [ ] AutoComplete : Issue with Chip closing
- [ ] AutoComplete : Dropdown menu not being at fixed position as scrolled through the screen

- [ ] ToggleSwitch : Add toggling transition for better UX.

**Steps to reproduce**
1. Run `yarn storybook` to open fuselage storybook
2. Click on the respective stories of `AutoComplete` and `ToggleSwitch`
3. Add chips in AutoComplete and click on the chip again.
4. Also scroll through the screen while the dropdown is switched on.

**Insert an image/gif/video**

[toggle-switch.webm](https://github.com/RocketChat/fuselage/assets/95426993/6d4d6ac4-04b9-42ad-9dea-8bb0e2b3d38d)

[AutoComplete.webm](https://github.com/RocketChat/fuselage/assets/95426993/118694f2-a30f-4113-85d0-b0003aa11fde)

vivekchavan14:
Is "yarn storybook" right command to run it?

vivekchavan14:
submitted a PR

umangutkarsh:
> submitted a PR

Hi @vivekchavan14 . Thanks for the PR. But I was already working on this.

vivekchavan14:
I thought it was a general issue, and i felt like i could solve it, thats why i solved the issue.

SySagar:
hey @umangutkarsh , did you try turning the inspect mode on and then scroll. The error you mentioed has been there in many such componets. So I think this is some sort of generalise issue.

juliajforesti:
Hey @umangutkarsh! 
Thanks for reporting these, we are working on fixing them.
[This PR](https://github.com/RocketChat/fuselage/pull/1755) fixes the behavior to remove selected item.

PR #1300, opened 2024-02-25
fixed autocompelete issue

vivekchavan14:
Made minor changes in handleSelect and handleRemove to fix Autocompelete issue

changes made : 

Changes in handleSelect:
1. Added a check to ensure that currentValue is truthy before proceeding.
2. Modified the logic to handle multiple selections and correctly update the value state based on whether it's an array or a single value.

Changes in handleRemove:
1. Added a removedValue variable to store the value of the removed item.
2. Utilized event.currentTarget.value to obtain the value of the removed item.
4. Updated the filtering logic to remove the item with the corresponding value from both the selected list and the value state.

issue: #1298

### DIFF
--- a/packages/fuselage/src/components/AutoComplete/AutoComplete.tsx
+++ b/packages/fuselage/src/components/AutoComplete/AutoComplete.tsx
@@ -77,14 +77,14 @@ export function AutoComplete({
   );
 
   const handleSelect = useEffectEvent(([currentValue]) => {
-    if (selected?.some((item) => item.value === currentValue)) {
-      hide();
-      return;
-    }
+    if (!currentValue) return; // Exit early if no value is selected
 
     if (multiple) {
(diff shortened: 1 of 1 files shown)


dougfabris:
Hi there, thanks for the contribution! 🚀 💯 

We're planning an entire replacement of all the components related to `AutoComplete` and all the `Select` components since we're focusing a lot on accessibility. Trying to maintain it became a nightmare, and because of that we're not going to accept any contributions related to these components at this time. 

---

SySagar:
hi @dougfabris since we are replacing these components which seems a breaking change to the codebase , what are we going to use as its replacement?

dougfabris:
@SySagar They are going to be replaced by brand new components with the same visual, same behavior but fully working and accessible. Because of that, we'll not receive small fixes turning the code hard to understand at this point

Issue #1301, opened 2024-02-26
Select Input logic error


SySagar:
**Description**
The placeholder option for select input shouldn't get selected

**Steps to reproduce**
1. Go to fuselage hosted link
2. Check the placeholder option in the select input.

**Insert an image/gif**
![Screenshot (35)](https://github.com/RocketChat/fuselage/assets/93787385/2ce718f1-f2af-4b73-9852-10c2ef37e833)

ShiruvatiNarasimha:
I can fix this issue

ritikpal1122:
Hey @SySagar assign this to me

sugalipravallikabai:
i taking this issue i will get back with resolution

sugalipravallikabai:
could you tell me the file name or folder name of that issue where it raised

PR #1302, opened 2024-02-26
fix(selectInput): temporary fix for select input with placeholder

SySagar:
fix(selectInput): temporary fix for select input with placeholder

Temporarily solves #1301 
## Further comments

This is a temporary fix and the actual change would require breaking change in codebase


SySagar:
hey @dougfabris , can you please suggest what further can I do in this pr

PR #1322, opened 2024-03-15
fix(fuselage): change PaginatedMultiSelectFiltered component to use PaginatedMultiSelect props

csuadev:
Change PaginatedMultiSelectFiltered component to use PaginatedMultiSelect props.


dougfabris (changes requested):
Please, check if my changes make sense to you.
Also if you want to generate a new version to be able to use it, you have to add a changeset. 
**Remember**: we only create changesets for fixes (patch) or feat (minor)

dougfabris (approved):

PR #1348, opened 2024-04-03
chore: Add `eslint-plugin-jsx-a11y` plugin

dougfabris:


packages/fuselage/src/components/Sidebar/SidebarBanner.tsx
```diff
@@ -30,13 +30,16 @@ export const SidebarBanner = ({
       {text && <div className='rcx-sidebar-banner--text'>{text}</div>}
       {description && (
         <div
+          role='link'
```
juliajforesti, line 33: is it really always a link?

packages/fuselage/src/components/SidebarV2/SidebarLink.tsx
```diff
@@ -18,6 +18,8 @@ export const SidebarLink = ({
   menu?: ReactNode;
 } & LinkHTMLAttributes<HTMLAnchorElement>) => (
   <a
+    role='link'
+    tabIndex={0}
```
juliajforesti, line 22: are these attributes needed here? the element is already `<a />` - which has role link and tabIndex 0 by default, isn't it?

juliajforesti (changes requested):

packages/fuselage/src/components/Sidebar/SidebarBanner.tsx
dougfabris: adding it conditionally

juliajforesti (approved):

PR #1383, opened 2024-05-24
fix(fuselage): remove `codeBlock` palette and add `palette` prop

juliajforesti:

- remove `codeBlock` palette from dark theme
- add `palette` prop for custom CSS palette


dougfabris (approved):
LGTM!

PR #1366, opened 2024-04-25
feat(fuselage): `Contextualbar` components size improvements

juliajforesti:
[SIDE-21](https://rocketchat.atlassian.net/browse/SIDE-21)
[SIDE-24](https://rocketchat.atlassian.net/browse/SIDE-24)
[SIDE-74](https://rocketchat.atlassian.net/browse/SIDE-74)

## Proposed changes (including videos or screenshots)

prev:
image 
new: 
image

## Further comments

[SIDE-21]: https://rocketchat.atlassian.net/browse/SIDE-21?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SIDE-24]: https://rocketchat.atlassian.net/browse/SIDE-24?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SIDE-54]: https://rocketchat.atlassian.net/browse/SIDE-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SIDE-74]: https://rocketchat.atlassian.net/browse/SIDE-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1404, opened 2024-06-21
fix(fuselage): move size changes into V2 components

juliajforesti:

The changes of the PRs below need to be moved into a V2 folder to support both versions until final deprecation
#1366 
#1369


dougfabris (approved):
follow your dreams

PR #1406, opened 2024-06-24
feat(fuselage-toastbar): Pause ToastBar timer while hovering

debdutdeb:


dougfabris (changes requested):
- Would be better to pause the interval while the user is hovering
- Since it changes the end user experience, it shouldn't be a `chore`

dougfabris (dismissed):

dougfabris (dismissed):

dougfabris (approved):

packages/fuselage-toastbar/src/ToastBarTimed.tsx
```diff
@@ -11,19 +11,30 @@ const ToastBarTimed = ({
...
 
   useEffect(() => {
-    const timeOut = setTimeout(() => {
-      dismissToastMessage(id);
-    }, time * 1000);
+    const initialTime = Date.now();
+    const interval = setInterval(() => {
+      if (shouldPause) {
+        return;
+      }
```
tassoevan, line 23: This condition is not going to work since the `setInterval`'s callback closure keeps a copy of the original `shouldPause` value.

https://react.dev/reference/react/useEffect#updating-state-based-on-previous-state-from-an-effect

packages/fuselage-toastbar/src/ToastBarTimed.tsx
```diff
@@ -11,19 +11,30 @@ const ToastBarTimed = ({
...
+    }, timeInSeconds);
+
+    return () => clearInterval(interval);
+  }, [shouldPause]);
 
   return (
     <ToastBar
       variant={type}
+      onMouseEnter={() => setShouldPause(true)}
+      onMouseLeave={() => setShouldPause(false)}
```
tassoevan, line 37: I guess we should prefer pointer events over mouse events.

tassoevan (changes requested):

tassoevan (approved):

PR #1409, opened 2024-07-17
feat(fuselage): `TableSelection` visual improvements 

dougfabris:

### DIFF
--- a/packages/fuselage/src/components/Table/TableSelection/TableSelectionButton.tsx
+++ b/packages/fuselage/src/components/Table/TableSelection/TableSelectionButton.tsx
@@ -1,10 +1,9 @@
 import type { ComponentProps } from 'react';
 import React from 'react';
 
-import type Box from '../../Box';
 import { Button } from '../../Button';
 
-type TableSelectionButtonProps = ComponentProps<typeof Box>;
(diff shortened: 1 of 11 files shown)


packages/fuselage/src/components/Table/TableSelection/TableSelectionButton.tsx
```diff
@@ -1,11 +1,11 @@
 import type { ComponentProps } from 'react';
 import React from 'react';
 
-import type Box from '../Box';
-import { Button } from '../Button';
+import type Box from '../../Box';
+import { Button } from '../../Button';
 
 type TableSelectionButtonProps = ComponentProps<typeof Box>;
```
juliajforesti, line 7: shouldn't it be `ComponentProps<typeof Button>`?

juliajforesti (changes requested):

packages/fuselage/src/components/Table/TableSelection/TableSelectionButton.tsx
dougfabris: it should

juliajforesti (approved):

PR #1415, opened 2024-07-24
feat(fuselage): Introduce `Callout` actions

dougfabris:

test


packages/fuselage/src/components/Callout/Callout.styles.scss
```diff
@@ -75,13 +75,35 @@ $callout-text-color: theme('callout-text-color', colors.font(default));
   &__wrapper {
     display: flex;
 
-    overflow: hidden;
```
juliajforesti, line 78: won't this be a problem? 👀

juliajforesti (changes requested):

packages/fuselage/src/components/Callout/Callout.styles.scss
dougfabris: gracias! 🙇

juliajforesti (approved):

juliajforesti (approved):

PR #1421, opened 2024-08-01
refactor(fuselage): Use JSX runtime instead of React namespace

tassoevan:

It makes Fuselage components use the JSX runtime transform and removes the synthetic default import of the `react` package.


dougfabris (approved):
LGTM!

PR #1443, opened 2024-09-04
fix(fuselage): Responsive CardControls  

### LINKED ISSUE #1440: Card: CardControls is not completely responsive
**Description**

The CardControls component is not responsive to mobile view causing issues in the Rocket Chat web app as mentioned [here](https://github.com/RocketChat/Rocket.Chat/issues/31614). Cards with multiple buttons should wrap within the container for responsiveness.

**Steps to reproduce**

1. Add multiple buttons to a card
2. Change to mobile view
3. You can see the unresponsive behaviour of the cards.

https://github.com/user-attachments/assets/a5a44dbe-418c-45fb-be81-f309c7cadd2e

Barrylimarti:

Added flex-wrap in the CSS of Cards for CardControl component. Things that have been kept in mind are:-
1. Wrap is not essential for horizontal cards in desktop view.
2. Wrap is enabled for horizontal cards in mobile view.
3. In every other case, flex-wrap is essential for responsive behaviour.
4. Made Carcontrol into a box component for better integration.

https://github.com/user-attachments/assets/ac6b0c39-c7a3-4dc8-ba7e-c592d18bbc4b

## Issue(s)
Closes #1440 [#31614 ](https://github.com/RocketChat/Rocket.Chat/issues/31614)

## Further comments
There are other ways to go around it but this felt appropriate and fixes the issue.

### DIFF
--- a/packages/fuselage/src/components/Card/Card.styles.scss
+++ b/packages/fuselage/src/components/Card/Card.styles.scss
@@ -83,4 +83,7 @@ $card-hero-padding: lengths.padding(28);
     display: flex;
     align-items: center;
   }
+  &__controls{
+    flex-wrap: wrap;
+  }
 }
--- a/packages/fuselage/src/components/Card/Card.styles.scss
+++ b/packages/fuselage/src/components/Card/Card.styles.scss
@@ -86,4 +86,10 @@ $card-hero-padding: lengths.padding(28);
   &__controls{
     flex-wrap: wrap;
   }
+  &__horizontal &__controls{
+    flex-wrap: nowrap;
+    &--wrap{
+      flex-wrap: wrap;
+    }
+  }
 }
--- a/packages/fuselage/src/components/Card/CardControls.tsx
+++ b/packages/fuselage/src/components/Card/CardControls.tsx
@@ -1,9 +1,13 @@
+import { useBreakpoints } from '@rocket.chat/fuselage-hooks';
 import Box from '../Box/Box';
 
 const CardControls = ({...props }) => {
+  const breakpoints = useBreakpoints();
+  const isMobile = !breakpoints.includes('sm');
   return (
     <Box
       rcx-card__controls
+      rcx-card__controls--wrap={isMobile}
       {...props}
     />
   )
--- a/packages/fuselage/src/components/Card/CardControls.tsx
+++ b/packages/fuselage/src/components/Card/CardControls.tsx
@@ -1,5 +1,12 @@
-const CardControls = ({ ...props }) => (
-  <div className='rcx-card__controls' {...props} />
-);
+import Box from '../Box/Box';
+
+const CardControls = ({...props }) => {
+  return (
+    <Box
+      rcx-card__controls
+      {...props}
+    />
+  )
+}
 
 export default CardControls;


packages/fuselage/src/components/Card/CardControls.tsx
```diff
@@ -1,5 +1,16 @@
...
+  const breakpoints = useBreakpoints();
+  const isMobile = !breakpoints.includes('sm');
+  return (
+    <Box
+      rcx-card__controls
+      rcx-card__controls--wrap={isMobile}
+      {...props}
+    />
+  )
+}
```
rique223, line 14:
```suggestion
import { useBreakpoints } from '@rocket.chat/fuselage-hooks';

import Box from '../Box/Box';

const CardControls = ({ ...props }) => {
  const breakpoints = useBreakpoints();
  const isMobile = !breakpoints.includes('sm');
  return (
    <Box rcx-card__controls rcx-card__controls--wrap={isMobile} {...props} />
  );
};
```

Same thing as the last PR, there are prettier issues with this snippet.

rique223 (changes requested):

packages/fuselage/src/components/Card/CardControls.tsx
```diff
@@ -1,5 +1,13 @@
...
+
+import Box from '../Box/Box';
+
+const CardControls = ({ ...props }) => {
+  const breakpoints = useBreakpoints();
+  const isMobile = !breakpoints.includes('sm');
+  return (
+    <Box rcx-card__controls rcx-card__controls--wrap={isMobile} {...props} />
+  );
+};
```
rique223, line 11:
```suggestion
import type { ReactNode, MouseEvent } from 'react';

type CardControlsProps = {
  children?: ReactNode;
  onClick?: (e?: MouseEvent<HTMLOrSVGElement>) => void;
};

const CardControls = ({ onClick, children }: CardControlsProps) => (
  <div className='rcx-card__controls' onClick={onClick} children={children} />
);

export default CardControls;
```

There is no need to change this component to box as you can do the conditional styling using the `on-breakpoint` mixin. Also, I've changed the `{...props}` as we should avoid this pattern (I know this is not necessarily part of you PR but it would be good if it could be added)

packages/fuselage/src/components/Card/Card.styles.scss
```diff
@@ -83,4 +83,16 @@ $card-hero-padding: lengths.padding(28);
...
+    flex-wrap: wrap;
+  }
+
+  &__horizontal &__controls {
+    flex-wrap: nowrap;
+
+    &--wrap {
+      flex-wrap: wrap;
+    }
+  }
```
rique223, line 97:
```suggestion
  &__horizontal &__controls {
    flex-wrap: wrap;

    @include on-breakpoint(sm) {
      flex-wrap: nowrap;
    }
  }
```
The mixin in question, also I've updated your logic to follow a 'mobile first' flow.

rique223 (changes requested):

packages/fuselage/src/components/Card/CardControls.tsx
Barrylimarti: Yes, this definitely looks the better way to handle this, I couldn't find the on-breakpoint mixin at that time, that's why chose to convert this to a box. And about the props, I will keep that in mind. Thanks for these suggestions.

Barrylimarti:
Hey Henrique! I just saw a repetition of code caused by previous commit. I fixed that in the latest commit. Now everything looks fine to me. Can you please check?

packages/fuselage/src/components/Card/CardControls.tsx
```diff
@@ -1,5 +1,12 @@
-const CardControls = ({ ...props }) => (
-  <div className='rcx-card__controls' {...props} />
+import type { ReactNode, MouseEvent } from 'react';
+
+type CardControlsProps = {
+  children?: ReactNode;
+  onClick?: (e?: MouseEvent<HTMLOrSVGElement>) => void;
```
rique223, line 5:
```suggestion
  onClick?: MouseEventHandler<HTMLDivElement>;
```

rique223 (changes requested):

packages/fuselage/src/components/Card/CardControls.tsx
Barrylimarti: Thanks for pointing this out too. A much better way. I have committed this change too.

Barrylimarti:
Hey @rique223! All the changes requested by you are done from my side. Let me know if you are looking for something else.

rique223 (approved):
Lgtm

packages/fuselage/src/components/Card/CardControls.tsx
```diff
@@ -1,5 +1,12 @@
...
-  <div className='rcx-card__controls' {...props} />
+import type { ReactNode, MouseEventHandler } from 'react';
+
+type CardControlsProps = {
+  children?: ReactNode;
+  onClick?: MouseEventHandler<HTMLDivElement>;
+};
+
+const CardControls = ({ onClick, children }: CardControlsProps) => (
+  <div className='rcx-card__controls' onClick={onClick} children={children} />
```
juliajforesti, line 9: Why change this component though? `props` were already being spread...
I would only add the prop types:

```suggestion
import { AllHTMLAttributes } from 'react';

const CardControls = (props: AllHTMLAttributes<HTMLDivElement>) => (
  <div className='rcx-card__controls' {...props} />
```

juliajforesti (changes requested):

packages/fuselage/src/components/Card/CardControls.tsx
rique223: This was my mistake 😅 . I thought it was the standard to use only the necessary props and avoid spreading props like this, but if the default should be the other way around no worries from me.

.changeset/purple-shirts-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/fuselage": patch
+---
+
+Fixes the responsiveness of the card component by adding flex wrap to card controls such that it can accomodate multiple items without breaking the responsive behaviour
```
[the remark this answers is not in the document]
Barrylimarti: Ohh okay got it! Will keep this in mind. Thanks for pointing this out.

packages/fuselage/src/components/Card/CardControls.tsx
Barrylimarti: Yes I also thought the same that we require only necessary props when Henrique proposed the change. But nevermind if you want we can go the other way around.

PR #1469, opened 2024-10-22
fix: resolved the issue of validating telephone number

### LINKED ISSUE #1332: TelephonicInput: requires regex for number , '+' and braces only
**Description**
Currently the telephonic input accepts anything i.e. strings, numbers, symbols like %,&,) etc. It should be made to accept only numbers with few paramters like + and ()

**Steps to reproduce**
1. Go to [fuselage components](https://rocketchat.github.io/fuselage/fuselage/master/?path=/docs/inputs-telephoneinput--default)

2. Type anything inside the text field

3. It should show it as invalid rather it doesn't shows anything

**Insert an image/gif**
Capture

devanshkansagra:
In this pull request I have added a regular expression which checks for the type of input whether it is braces or '+' or numbers

[Screencast from 22-10-24 09:09:29 PM IST.webm](https://github.com/user-attachments/assets/3f0cb0b8-32f9-44e2-84f6-0821f84536f8)

## Issue(s)

Closes: #1332


devanshkansagra:
Hey @ggazzo and other maintainers, could you please review this pull request so that if there are any changes to be made I will follow up

PR #1470, opened 2024-10-22
feat(fuselage)!: Lighter `Accordion` and `AccordionItem` components

tassoevan:

It removes an obsolete and not accessible toggle switch in `AccordionItem` and eases the internal usage of `Box` to
improve rendering performance.

Additionally, it adds a new `StylingBox` component that can be used as a wrapper for components that accept styling
props but don't need the weight of the `Box` component prop handling internally.

Also, it adds a new `cx` and `cxx` helpers to compose class names.

## Further comments
An important tweak was made in Jest's setup, mocking `useUniqueId` calls to return deterministic IDs, allowing snapshot matching.

### DIFF
--- a/.changeset/soft-islands-decide.md
+++ b/.changeset/soft-islands-decide.md
@@ -1,0 +1,13 @@
+---
+"@rocket.chat/fuselage": minor
+---
+
+Simplifies `Accordion` and `AccordionItem`
+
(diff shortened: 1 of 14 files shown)


dougfabris (approved):
If we had the world ☮️, I think it would be something like this

PR #1492, opened 2024-12-02
fix: dropdown menu not fully visible in constrained containers

abhinavkrin:
This PR resolves the issue where dropdown menus are clipped or not fully visible inside containers with constrained height or overflow properties. The menus are now fully visible and accessible

#### before 
image

#### after
Screenshot 2024-12-02 at 6 31 58 PM

Fixes an issue where dropdown menus were clipped or hidden inside constrained containers.

## Further comments

[CORE-760](https://rocketchat.atlassian.net/browse/CORE-760)

[CORE-760]: https://rocketchat.atlassian.net/browse/CORE-760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/clever-planets-vanish.md
+++ b/.changeset/clever-planets-vanish.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/fuselage': patch
+---
+
+Fixes an issue where dropdown menus were clipped or hidden inside constrained containers.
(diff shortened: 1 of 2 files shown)


dougfabris:
@abhinavkrin this is not the behavior we always want for this component. In order to fix the issue you have you should use the `detached` property when using `Menu V2`

PR #1531, opened 2025-01-21
fix(fuselage-toastbar): React 18 compatibility

tassoevan:

It handles the fact that React 18's Strict Mode fires effects twice, which breaks Fuselage's toast bar portal.


dougfabris (changes requested):
for some reason these changes blocks the browser when tries to start the storybook and doesn't render anything, neither issues on console

dougfabris (approved):

PR #1533, opened 2025-01-22
fix(fuselage): `SelectAria` id prop overwriting

juliajforesti:

After `react-aria` bump the `id` prop was being overwritten causing an infinite loop


dougfabris (approved):
LGTM!

PR #1559, opened 2025-02-12
feat: replace `useLayoutEffect` in favor of `useInsertionEffect`

juliajforesti:

> useInsertionEffect allows inserting elements into the DOM before any layout Effects fire.
It's indicated for inserting styles before any Effects fire that may need to read layout.

ref: https://react.dev/reference/react/useInsertionEffect

This PR replaces `useLayoutEffect` in favor of `useInsertionEffect` in Box component's `useStyle` hook and in `styled` package.

### DIFF
--- a/packages/styled/src/styled.ts
+++ b/packages/styled/src/styled.ts
@@ -19,7 +19,7 @@ import {
   forwardRef,
   Fragment,
   useDebugValue,
-  useLayoutEffect,
+  useInsertionEffect,
 } from 'react';
 
 export const attachClassName = <P extends { className?: string }>(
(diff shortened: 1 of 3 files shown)


packages/styled/src/styled.ts
```diff
@@ -19,7 +19,8 @@ import {
   forwardRef,
   Fragment,
   useDebugValue,
-  useLayoutEffect,
+  useInsertionEffect,
+  // useLayoutEffect,
```
dougfabris, line 23:
```suggestion

```

dougfabris (changes requested):

packages/styled/src/styled.ts
juliajforesti: oops, thanks

dougfabris (approved):

PR #1564, opened 2025-02-18
fix(fuselage): Missing track on `Slider` component

rique223:

There was a rendering issue with the SliderTrack component in Chromium-based browsers. The component was using the CSS attr() function with an invalid attribute but without specifying fallback values or attribute types. While this edge case should return an empty string according to the CSS specification, Chromium browsers weren't handling it correctly, causing the track to fail to render. This has been fixed by explicitly setting an empty string ('') instead of relying on attr().

Before:
before_attr_bug

After:
after_attr_bug

### DIFF
--- a/.changeset/grumpy-mice-draw.md
+++ b/.changeset/grumpy-mice-draw.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/fuselage": patch
+---
+
+Fixes the missing track on the Slider component to ensure proper visual representation and functionality.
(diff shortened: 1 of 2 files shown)


juliajforesti (changes requested):
Add a changeset and you're good to go

.changeset/grumpy-mice-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rocket.chat/fuselage": patch
+---
+
+Fixes the missing track on the Slider component to ensure proper visual representation and functionality.
```
juliajforesti, line 5: On fuselage the pattern of the changeset is the same as the PR title:
```suggestion
fix(fuselage): Missing track on `Slider` component
```

juliajforesti (changes requested):

PR #1595, opened 2025-03-25
refactor: use `useSyncExternalStore` in favor of `useState` in `useXyzStorage` hooks

gabriellsh:

### DIFF
--- a/packages/fuselage-hooks/src/useStorage.ts
+++ b/packages/fuselage-hooks/src/useStorage.ts
@@ -1,6 +1,6 @@
 import { Emitter } from '@rocket.chat/emitter';
-import type { Dispatch, SetStateAction } from 'react';
-import { useRef, useCallback, useSyncExternalStore } from 'react';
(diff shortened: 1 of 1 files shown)


packages/fuselage-hooks/src/useStorage.ts
```diff
@@ -24,49 +37,56 @@ function makeStorageHook(
     const initialValueRef = useRef(initialValue);
     initialValueRef.current = initialValue;
 
-    const [storedValue, setStoredValue] = useState<T>(() => {
-      if (!storage) {
-        return initialValue;
-      }
+    const valueRef = useRef(getStorageItem(getKey(key), initialValue, storage));
```
tassoevan, line 40: During re-renders (caused by external reasons), `getStorageItem` will be called again.

tassoevan (changes requested):

tassoevan (approved):

PR #1622, opened 2025-04-15
feat(draggable-core): Create package

gabriellsh:


dougfabris:
why `@rocket.chat/draggable-core` it seems weird to me since we don't use this name pattern here. Why not `@rocket.chat/fuselage-draggable` or `@rocket.chat/draggable`?

PR #1632, opened 2025-05-07
fix(fuselage): `MenuV2` not handling focus properly 

yash-rajpal:


.changeset/loose-teams-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/fuselage': patch
+---
+
+Improves MenuV2 accessibility
```
juliajforesti, line 5: the pattern in this repo is to replicate the PR title:
```suggestion
fix(fuselage): `MenuV2` not handling focus properly 
```

packages/fuselage/src/components/Menu/V2/MenuPopover.tsx
juliajforesti, line 1: Why are you not using `DropdownDesktop / Mobile` here anymore?
You could make the changes needed in those components, since I don't see them being used elsewhere.

juliajforesti (dismissed):

packages/fuselage/src/components/Menu/V2/MenuPopover.tsx
dougfabris: they are used elsewhere that's why we didn't changed them

yash-rajpal: `DropDown` component is being used in a few places on the core repo, I'll create a task on our backlog to track this, and replace usage of this DropDown component in core. Thanks Julia!

juliajforesti: Indeed they are, but would it be a breaking change? I don't see too much of a diference... You're just not passing the expected style with `visibility` property on `DropdownDesktop` -- whose implementation seems odd (`(style as any).visibility`).

If no `style` is given, it should not break! I'd suggest changing to:
```
{style?.visibility === 'hidden' ? null : children}
```

please check if this would work.

dougfabris: in order to use `DropdownDesktop` or `DropdownMobile` we would need to pass `state` prop, because we need the `Popover` wrapping the `Tile`, and this will be a breaking change. That's why for now we're going to skip its usage

packages/fuselage/src/components/Menu/V2/MenuPopover.tsx
```diff
@@ -1,12 +1,11 @@
...
-import { usePopover } from 'react-aria';
 import type { OverlayTriggerState } from 'react-stately';
 
-import { DropdownDesktop } from '../../Dropdown/DropdownDesktop';
-import { DropdownMobile } from '../../Dropdown/DropdownMobile';
+import Box from '../../Box';
+import { Popover } from '../../Popover';
+import Tile from '../../Tile';
 
 interface PopoverProps extends Omit<AriaPopoverProps, 'popoverRef'> {
```
dougfabris, line 10:
```suggestion
interface MenuPopoverProps extends Omit<AriaPopoverProps, 'popoverRef'> {
```

packages/fuselage/src/components/Menu/V2/MenuPopover.tsx
```diff
@@ -21,31 +20,48 @@ function MenuPopover({
   maxWidth,
   ...props
 }: PopoverProps) {
```
dougfabris, line 22:
```suggestion
}: MenuPopoverProps) {
```

dougfabris (changes requested):

packages/fuselage/src/components/Menu/V2/MenuPopover.tsx
juliajforesti: @dougfabris @yash-rajpal I've commited the changes I was suggesting so it's more clear, as I don't believe it's a breaking change. Let me know if you agree :)

PR #1642, opened 2025-05-15
fix: 1.1.1 Changed tag from label to div in ToggleSwitch

ergot-rp:


dougfabris (changes requested):
this change will break the component, we need to think about a way to fix it properly

ivans-netto:
I've spoke with Henriques and removing the <label> wrapper from inputs will have other implications through out [Rocket.Chat](https://open.rocket.chat/#) software. The front-end team will need to decide what can be done in this case. Expect a reply until next week.

PR #1754, opened 2025-09-05
fix(fuselage)!: Bundling size

tassoevan:


packages/fuselage/src/components/Skeleton/Skeleton.tsx
```diff
@@ -1,7 +1,7 @@
-import { AllHTMLAttributes } from 'react';
+import type { AllHTMLAttributes } from 'react';
 
 import { cx, cxx } from '../../helpers/composeClassNames';
-import { StylingBox, StylingBoxProps } from '../Box';
+import { StylingBox, type StylingBoxProps } from '../Box';
```
dougfabris, line 4: there's no lint rule preventing us for this such thing?

dougfabris (approved):
`es-check`

packages/fuselage/src/components/Skeleton/Skeleton.tsx
tassoevan: Enabling the `@typescript-eslint/consistent-type-imports` rule will do the trick.

PR #1756, opened 2025-09-15
fix(fuselage): AutoComplete selected items not being updated when value changes

aleksandernsilva:
Previously, selected items in the AutoComplete component could remain visible even when the current value no longer contained them. This PR adds a validation step to ensure consistency between the value and the selected items, removing any invalid ones.

## Issue(s)
[CTZ-336](https://rocketchat.atlassian.net/browse/CTZ-336)

## Further comments

[CTZ-336]: https://rocketchat.atlassian.net/browse/CTZ-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/green-walls-kick.md
+++ b/.changeset/green-walls-kick.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/fuselage": patch
+---
+
+fix: AutoComplete selected items not being updated when value changes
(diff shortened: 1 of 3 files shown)


juliajforesti (approved):
Although we agree that using the `useEffect` to handle the validation is not the best practice, we are accepting this for now, because of the way `AutoComplete` is built.

PR #1811, opened 2026-01-01
fix(a11y): add dev warning for buttons missing accessible names

### LINKED ISSUE #1810: [Improvement]: Add development warning for icon-only Buttons missing accessible labels
### Description
In the current `Button` component, it is possible to implement icon-only buttons (or square buttons) without providing an `aria-label` or `aria-labelledby`. 

Example of inaccessible code currently allowed:
```tsx
<Button icon="edit" />
<Button square><Icon name="plus" /></Button>
```
This results in screen readers announcing only the word "button," which violates WCAG 2.1 (Success Criterion 4.1.2: Name, Role, Value).

### Proposed Solution
Add a development-time check (process.env.NODE_ENV !== 'production') within the Button component. If the button lacks visible text and also lacks an ARIA label, a console.warn should be triggered to guide the developer.

### Why this matters
Fuselage is the foundation for Rocket.Chat's UI. Ensuring accessibility at the component level prevents accessibility regressions across the entire ecosystem without introducing breaking changes to the API.

I have a local implementation ready and would like to submit a PR for this.

Kalpana-chavhan:
### What this PR does
This PR adds a non-breaking accessibility check to the `Button` component that warns developers when a button is visually "empty" (icon-only, loading-only, or square) and lacks an ARIA label.

### Why is this needed?
To ensure WCAG 2.1 compliance. Icon-only buttons are unusable for screen reader users unless an `aria-label` is provided. This warning helps developers catch these omissions during development.

### Changes
- Updated `Button.tsx` to include `Children` from React.
- Added logic to detect "visual-only" buttons.
- Included a `console.warn` wrapped in a production-environment check.

### Testing
1. Render `<Button icon="edit" />` -> Console should show warning.
2. Render `<Button icon="edit" aria-label="Edit" />` -> No warning.
3. Render `<Button>Save</Button>` -> No warning.

closes #1810

### DIFF
--- a/packages/fuselage/src/components/Button/Button.tsx
+++ b/packages/fuselage/src/components/Button/Button.tsx
@@ -1,5 +1,5 @@
 import type { AllHTMLAttributes } from 'react';
-import { forwardRef, useMemo } from 'react';
+import { forwardRef, useMemo, Children } from 'react';
 
 import { Box, type BoxProps } from '../Box';
 import { Icon, type IconProps } from '../Icon';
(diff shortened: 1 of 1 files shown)


dougfabris:
Hi there, thanks for the contribution 🚀 

We're not interested on this because it won't add much value in our codebase, because of that I'm closing this one

---

PR #1832, opened 2026-01-13
feat(fuselage): Increase `MessageBlock` margins when placed directly bellow `MessageHeader`

gabriellsh:
Increased the margin of the `MessageBlock` component when it's directly below the message header.
#### Before:
Screenshot 2026-01-13 at 10 17 25

#### After:
Screenshot 2026-01-13 at 10 17 38

## Issue(s)

[VGA-136](https://rocketchat.atlassian.net/browse/VGA-136)
## Further comments

[VGA-136]: https://rocketchat.atlassian.net/browse/VGA-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/fuselage/src/components/Message/Messages.stories.tsx
+++ b/packages/fuselage/src/components/Message/Messages.stories.tsx
@@ -6,12 +6,47 @@ import { Box } from '../Box';
 
 import Message from './Message';
 import MessageBlock from './MessageBlock';
+import MessageBody from './MessageBody';
+import MessageContainer from './MessageContainer';
 import { MessageDivider } from './MessageDivider';
 import MessageEmoji from './MessageEmoji';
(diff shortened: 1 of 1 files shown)


packages/fuselage/src/components/Message/Messages.stories.tsx
dougfabris: Can we avoid using `Parent.Child` notation, there are [some efforts](https://github.com/RocketChat/fuselage/pull/1805) to get away from this

gabriellsh: I think this would be out of scope for this one though, right?

PR #1875, opened 2026-02-28
test: add missing type attribute and behavior tests for Button component

### LINKED ISSUE #1813: feat(fuselage): ensure Button defaults to type="button" to prevent accidental form submissions  Body:
### Description
Currently, HTML `<button>` elements default to `type="submit"` when placed inside a `<form>`. In a large application like Rocket.Chat, using a Fuselage `<Button>` for actions like "Cancel" or "Close" inside a form context can lead to unintended submissions if the `type` is not explicitly set.

### Proposed Solution
Modify the `Button` component to default to `type="button"` while still allowing developers to explicitly override it with `type="submit"` or `type="reset"`.

### Why this matters
This is a defensive programming pattern that prevents subtle UX bugs. By ensuring buttons are "neutral" by default, we improve the reliability of forms across the Rocket.Chat ecosystem.

### Changes
- Set default `type = 'button'` in the component props.
- Ensure the `type` prop is correctly forwarded to the underlying `Box` or `button` element.
- Maintain support for `type="submit"` when explicitly provided.

I have analyzed the current `Button.tsx` implementation and submitted a PR with this safer default.

Peersheik:
Adds missing unit tests for the Button component to verify:

1. Button defaults to type="button" to prevent accidental form submissions
2. Button allows overriding type attribute when explicitly passed
3. Button renders as anchor tag when is="a" prop is passed

The existing tests only checked rendering and accessibility (a11y).
There were no tests verifying the Button's type attribute behavior,
which is critical to prevent accidental form submissions in apps
using Fuselage components.

Issue(s):
Closes #1813

Further comments:
This PR adds test coverage for the default type="button" behavior 
discussed in #1813. The Button component already correctly sets 
type="button" as default, but there were no tests to verify this 
behavior and prevent future regressions.

Tests added:
- Verifies Button renders with type="button" by default
- Verifies type attribute can be overridden when explicitly passed
- Verifies Button renders as anchor when is="a" prop is passed

### DIFF
--- a/packages/fuselage/src/components/Button/Button.spec.tsx
+++ b/packages/fuselage/src/components/Button/Button.spec.tsx
@@ -52,4 +52,22 @@ describe('[IconButton Component]', () => {
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
+
+  it('should default to type="button" to prevent accidental form submissions', () => {
+    const { getByRole } = render(<Default />);
+    const button = getByRole('button');
+    expect(button).toHaveAttribute('type', 'button');
+  });
+
+  it('should allow overriding type attribute when explicitly passed', () => {
+    const { getByRole } = render(<Default type='submit' />);
+    const button = getByRole('button');
+    expect(button).toHaveAttribute('type', 'submit');
+  });
+
+  it('should render as anchor tag when is="a" prop is passed', () => {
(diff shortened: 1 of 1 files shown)


dougfabris:
Hi there, thanks for the contribution 🚀 

We're not interested on this because it won't add much value in our codebase, because of that I'm closing this one

---

PR #1894, opened 2026-03-06
fix(fuselage): Improve form input accessibility (A11y)

Harshit2405-2004:
## What does this PR do?
Fixes PasswordInput a11y violations by adding ARIA attributes to the toggle icon.

## Motivation
Password inputs lack proper ARIA labels and roles, causing screen readers to fail axe validation.

## Changes Made
- Added aria-label, role='button', tabIndex, and onKeyDown handlers to the password toggle icon.


dougfabris (changes requested):
We could use `IconButton` as add-on instead

dougfabris:
The proper variant is indeed `tiny` in order to match the size of input, but it feels so small. 
I will need to check with design team before we move forward!

Harshit2405-2004:
@dougfabris I'll surely fix according to the requirements. so, you can definitely let me know about the same. I can update that appropriately .

PR #1933, opened 2026-04-13
feat(css-in-js): Allow passing a target `document` to rule attacher

gabriellsh:
Today `css-in-js` directly references the `window.document` property in the style attachers. This is fine for our current use, but in some situations (such as portals inside popup windows) this can cause a mismatch between the document that the element is being rendered, and the document where the styles are being inserted.

This PR introduces an option to pass `document` as parameter, allowing for different target documents to be used.

## Issue(s)

[PRD-323](https://rocketchat.atlassian.net/browse/PRD-323)
## Further comments

[PRD-323]: https://rocketchat.atlassian.net/browse/PRD-323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/smooth-deers-burn.md
+++ b/.changeset/smooth-deers-burn.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/css-in-js": minor
+---
+
+feat(css-in-js): Allow passing a target `document` to rule attacher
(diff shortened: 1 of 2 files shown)


gabriellsh:
@coderabbitai review

tassoevan (approved):

PR #1765, opened 2025-10-14
chore(fuselage): Sass overhaul

tassoevan:
## Proposed changes (including videos or screenshots)

It removes the conditions that generate compilation warnings from the Sass compiler during `@rocket.chat/fuselage` build. Additionally, it eliminates some kilobytes of accidentally duplicated CSS rules.

## Issue(s)

Closes #1665

[ARCH-2122](https://rocketchat.atlassian.net/browse/ARCH-2122)

## Further comments

This pull request refactors SCSS files across multiple Fuselage components to standardize the way style modules, mixins, and variables are imported and used. The main goals are to simplify and unify import syntax, switch to namespaced or wildcard imports, and remove redundant prefixes from mixin and function calls. Additionally, some code organization improvements and minor bug fixes are included.

**Standardization and Refactoring of SCSS Imports and Mixin Usage:**

[...]

PR #1961, opened 2026-05-07
fix(fuselage): issues not caught by visual regression testing after Sass rewrite

tassoevan:

It fixes issues in `Avatar`, `Chip`, and `<a>` elements.

## Further comments

@gabriellsh suggested my visual regression test was botched due to pixel diff tolerance getting bamboozled in large snapshots; he was right.

Introduced here https://github.com/RocketChat/fuselage/pull/1765

### DIFF
--- a/.changeset/ten-cats-occur.md
+++ b/.changeset/ten-cats-occur.md
@@ -1,0 +1,5 @@
+---
+"@rocket.chat/fuselage": patch
+---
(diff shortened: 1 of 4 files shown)


gabriellsh (dismissed):
Thanks!

gabriellsh (approved):

PR #1982, opened 2026-05-22
fix(fuselage): autocomplete overlay on MultiSelectFiltered component

nazabucciarelli:
This PR fixes a UI bug where Chrome's native autocomplete/history dropdown overlaps and obscures the custom options list in the MultiSelectFiltered component (specifically observed when searching for users to create a Direct Message in Rocket.Chat).

## Root Cause Analysis & Investigation
After deep-diving into the component's history and the recent refactoring, I discovered why this bug suddenly appeared:

The "Before" state (Why Chrome ignored it previously): In older versions of `MultiSelect.tsx`, attributes like `name` and `id` were being spread onto the parent wrapper (<Box is="div">). The actual <input> rendered by the anchor was left "anonymous" (without a name attribute). Because it lacked a recognizable name, Chrome's aggressive heuristic engine ignored it and did not attempt to autofill it.

The "After" state (Why the bug appeared): During a recent refactoring to improve accessibility and HTML standards, the id and name attributes were explicitly extracted and correctly passed down to the actual <input> element inside renderAnchor. The PR in which this was done is https://github.com/RocketChat/fuselage/pull/1940

## The small fix
I added `autoComplete="off"` directly to the <Input> element inside `MultiSelectFilteredAnchor.tsx` because an input used strictly as an internal search/filter for a combobox dropdown should never trigger browser autofill. Hardcoding this protection at the base level ensures the component is inherently safe, rather than relying on the parent to remember to pass the prop.

## Issue(s)

[CORE-2205 [Regression] Persistent history bar remains visible in user search](https://rocketchat.atlassian.net/browse/CORE-2205)


gabriellsh (approved):

gabriellsh (changes requested):
Actually, I think we must make use of `aria-autocomplete` in it's place. This component is, in a certain way, completing the user's input, or as described by [mdn's docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-autocomplete)

> The aria-autocomplete attribute indicates whether inputting text could trigger display of one or more predictions of the user's intended value for a [combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role), [searchbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/searchbox_role), or [textbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/textbox_role) and specifies how predictions will be presented if they are made.

I'm still uncertain but I'm guessing this property has nothing to do with the default "autocomplete" property. I'm doing some resarch and will get back with a proper conclusion.

gabriellsh (approved):
My interpretation here https://github.com/RocketChat/fuselage/pull/1982#pullrequestreview-4364608165 was correct, but it seems like we're also missing some other stuff to make it properly accessible. Since this would require more extensive changes and testing, let's move with the fix as it is.

PR #2043, opened 2026-06-29
fix(fuselage): InputBox not stretching to the maximum available width

nazabucciarelli:

`InputBox`, and everything built on it (`TextInput`, `UrlInput`, `EmailInput`, `NumberInput`, `TelephoneInput`, `TextAreaInput`, etc.) collapses to its `min-width` instead of filling the available space when it is rendered as a direct child of a flex‑column container (e.g. a fuselage `Field`, which is `display: flex; flex-direction: column; align-items: stretch`) rather than inside a `FieldRow`.

**Root cause**: v0.79.1 (https://github.com/RocketChat/fuselage/pull/2024) added `width: 0` to the shared `.rcx-input-box` rule. A definite `width` disables the cross‑axis `stretch` of a flex column item, so the input stops growing to the container width and falls back to `min-width: 8rem`. In v0.79.0 and earlier the rule had no `width`, so it stretched correctly.

**Fix**: Use a main‑axis‑only basis instead of a definite width:

**Where it shows**: in Rocket.Chat admin settings, every `relativeUrl` setting (e.g. the OAuth "Callback URL" fields) renders a readonly `UrlInput` directly inside a `Field`, and appears collapsed to ~128px with the URL truncated, while sibling text/secret fields (which sit inside a `FieldRow`) look fine.

- **The defect is in the component's sizing contract, not in one call site.** `width: 0` breaks `InputBox` for *any* addon‑less input rendered in a stretch (column/block) container, which is a valid usage. The `relativeUrl` setting (the one used in `apps/meteor/server/settings/oauth.ts`) is just the first place it became visible: `TextInput`, `EmailInput`, etc. collapse identically there, they are the same component with a different `type`. Fixing it in Rocket.Chat main repo would patch one symptom and leave the latent bug for every other current and future consumer.

## Issue(s)
[CORE-2348 [Regression] OAuth callback URL fields render too narrow in admin settings](https://rocketchat.atlassian.net/browse/CORE-2348)

### DIFF
--- a/.changeset/cold-knives-post.md
+++ b/.changeset/cold-knives-post.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/fuselage': patch
+---
+
+fix(fuselage): InputBox not stretching to the maximum available width
--- a/packages/fuselage/src/components/InputBox/InputBox.styles.scss
+++ b/packages/fuselage/src/components/InputBox/InputBox.styles.scss
@@ -76,6 +76,7 @@ $input-border-radius: theme(
   flex: 1 0 auto;
 
   flex-basis: lengths.size(none);
+
   min-width: lengths.size(128);
 
   user-select: initial;
--- a/packages/fuselage/src/components/InputBox/InputBox.styles.scss
+++ b/packages/fuselage/src/components/InputBox/InputBox.styles.scss
@@ -75,7 +75,7 @@ $input-border-radius: theme(
 
   flex: 1 0 auto;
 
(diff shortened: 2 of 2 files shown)


nazabucciarelli:
We decided to fix it on the main repo, as the defined css `width` property is needed and shouldn't be changed because it was added to fix a Firefox visual issue. Closing this PR in favor of https://github.com/RocketChat/Rocket.Chat/pull/41105

PR #2049, opened 2026-06-30
refactor: Get rid of `forwardRef`

tassoevan:

It removes redundant usage of `forwardRef`.

## Further comments

It's not expected compile-time changes with this changes. Most of the impact might be perceived as a minimal runtime perfomance improvement.

### DIFF
--- a/packages/layout/layout.api.md
+++ b/packages/layout/layout.api.md
@@ -16,7 +16,6 @@ import { ForwardRefExoticComponent } from 'react';
 import { FunctionComponent } from 'react';
 import { HTMLAttributes } from 'react';
 import { Icon } from '@rocket.chat/fuselage';
-import { IconProps } from '@rocket.chat/fuselage';
 import { JSX } from 'react';
 import { MarginsProps } from '@rocket.chat/fuselage';
 import { MemoExoticComponent } from 'react';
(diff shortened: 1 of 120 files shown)


packages/layout/layout.api.md
```diff
@@ -80,38 +78,30 @@ type DarkModeProviderProps = {
 // @public (undocumented)
 const Description: FunctionComponent<ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement>>;
 
+// Warning: (ae-forgotten-export) The symbol "Form_2" needs to be exported by the entry point index.d.ts
```
dougfabris, line 81: have we forgot to export some components? Asking because of these warnings

dougfabris (changes requested):
found some things: 
- `yarn install` revert some changes you did to `...api.md`
- we still need the `isForwardRefType` helper?
- since you replaced all `addon` in favor of `endAddon` can we drop the deprecated prop in this pull request?

packages/layout/layout.api.md
tassoevan: Those are hard to fix, and I haven't found a good solution. The problem here is that there is a `function Form(...) { ... }` and a exported `Form` that is `memo(Form)`. Exporting both to appease API Extractor feels idiotic.

tassoevan:
> since you replaced all addon in favor of endAddon can we drop the deprecated prop in this pull request?

coming in a following PR

dougfabris (approved):

PR #2052, opened 2026-07-01
feat(fuselage): extract controlled AudioPlayerControls from AudioPlayer

ggazzo:
## Why

`AudioPlayer` owns its own `<audio>` element and playback state internally, so it can't be driven by an external/shared media element. That makes cases like a persistent player (audio that keeps playing while the UI that hosts it unmounts, e.g. navigating between rooms) impossible to build on top of it — consumers currently have to re-implement the whole control layout.

## What

Extracts the presentational part of `AudioPlayer` into a new, fully **controlled** `AudioPlayerControls` component and exports it:

- `AudioPlayerControls` renders play/pause, seek, elapsed time, playback speed and the optional download button. It owns **no** `<audio>` element and **no** playback state — everything is driven by props (`isPlaying`, `currentTime`, `durationTime`, `playbackSpeed`, `onTogglePlay`, `onSeek`, `onChangePlaybackSpeed`, download props and labels).
- `AudioPlayer` keeps its previous self-contained behavior (creates the `<audio>`, manages state) and now composes `AudioPlayerControls` internally.

Consumers that need to drive a shared/external element use `AudioPlayerControls` directly; everyone already using `AudioPlayer` is unaffected.

## Backward compatibility

Non-breaking. `AudioPlayer`'s public props and rendered output are unchanged — it just delegates its inner layout to `AudioPlayerControls`.

## Notes

Added a `Controls` story demonstrating the controlled usage. Changeset included (`@rocket.chat/fuselage`: minor).

### DIFF
--- a/.changeset/audio-player-controls.md
+++ b/.changeset/audio-player-controls.md
@@ -2,4 +2,4 @@
 '@rocket.chat/fuselage': minor
 ---
 
-Extracted the presentational, fully controlled `AudioPlayerControls` component from `AudioPlayer` and exported it. It renders the play/pause, seek, elapsed time, playback speed and optional download controls without owning an `<audio>` element or any playback state, so it can be driven by an external/shared media element. `AudioPlayer` keeps its previous self-contained behavior and now composes `AudioPlayerControls` internally, so existing usage is unchanged.
+feat(fuselage): extract controlled AudioPlayerControls from AudioPlayer
--- a/.changeset/audio-player-controls.md
+++ b/.changeset/audio-player-controls.md
@@ -1,0 +1,5 @@
+---
+'@rocket.chat/fuselage': minor
+---
+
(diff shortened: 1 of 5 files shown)


.changeset/audio-player-controls.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/fuselage': minor
+---
+
+Extracted the presentational, fully controlled `AudioPlayerControls` component from `AudioPlayer` and exported it. It renders the play/pause, seek, elapsed time, playback speed and optional download controls without owning an `<audio>` element or any playback state, so it can be driven by an external/shared media element. `AudioPlayer` keeps its previous self-contained behavior and now composes `AudioPlayerControls` internally, so existing usage is unchanged.
```
juliajforesti: pattern here is to use the same PR title to the changeset
```suggestion
feat(fuselage): extract controlled AudioPlayerControls from AudioPlayer
```

juliajforesti (changes requested):

juliajforesti (approved):

PR #2094, opened 2026-07-16
feat(fuselage): opt-in atomic styling path for Box

ggazzo:
## What

Adds an **experimental, opt-in** atomic styling path for `Box`, so we can
compare it against the current per-combination approach on real screens.

### Today (per-combination)
All styling props on a `Box` are merged into one CSS blob, hashed, and injected
as a single class. A declaration like `display: flex` is duplicated inside every
combination class, so the stylesheet grows with the number of distinct **combinations**.

### New (atomic)
Emits **one class per styling prop**. `attachRules` already dedupes identical
rules per document and ref-counts them, so shared declarations resolve to a
single reused rule — the stylesheet grows with the number of distinct
**declarations** instead.

## How to compare

Toggle in the browser console, then reload (flag is read once at module load):

```js
localStorage.setItem('fuselage-styling', 'atomic') // atomic (experimental)
localStorage.removeItem('fuselage-styling')        // default (per-combination)
```

Default is unchanged production behavior; atomic is strictly opt-in.

## Correctness

- `useStylingProps` selects the path once at module load, so the render branch
  is stable and the two paths never mix their hooks.
- A spec asserts the atomic path produces the **exact same declarations** as the
  merged path (same visuals), plus one-class-per-prop and cross-Box dedup.

## Notes

- Draft — for A/B comparison, not merge-ready.
- No changeset yet (experimental).
- Follow-up if it wins: move the same per-prop resolution to build time
  (precomputed classNames, drop the runtime `css()` call).

### DIFF
--- a/packages/fuselage/.storybook/main.ts
+++ b/packages/fuselage/.storybook/main.ts
@@ -36,7 +36,15 @@ const config: StorybookConfig = {
                 ],
               ],
               plugins: [
-                [boxPlugin, { styleProps, resolve: resolveBox, inject: true }],
+                [
+                  boxPlugin,
+                  {
+                    styleProps,
+                    resolve: resolveBox,
+                    inject: true,
+                    keepProps: true,
+                  },
+                ],
               ],
             },
           },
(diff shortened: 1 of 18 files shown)


ggazzo:
Superseded — split into two stacked PRs: #2097 (opt-in atomic styling path → main) and #2098 (build-time compiler PoC, stacked on #2097).

PR #2100, opened 2026-07-16
feat(fuselage)!: Restrict border radius and border width token names

tassoevan:

`Box`'s border width and radius props now take border token names instead of arbitrary CSS lengths or `x`-scale measures:

- `borderWidth`, `borderBlockWidth`, `borderBlockStartWidth`, `borderBlockEndWidth`, `borderInlineWidth`, `borderInlineStartWidth` and `borderInlineEndWidth` accept `none`, `default`, `medium`, `large` and `extra-large`.
- `borderRadius`, `borderStartStartRadius`, `borderStartEndRadius`, `borderEndStartRadius` and `borderEndEndRadius` accept `none`, `default`, `small`, `medium`, `large`, `extra-large` and `full`.

The values are resolved from the `border` tokens in `@rocket.chat/fuselage-tokens`, and the `BorderWidth` and `BorderRadius` types are exported. Anything outside that vocabulary is now a type error, and resolves to `undefined` at runtime instead of being passed through as CSS.

Other changes:

- Adds the missing `extra-large` (8px) border width token.
- `lengths.border-width()` in SCSS drops its numeric branch (`1`, `2`, `4`) and resolves token names only.
- `fuselage-tokens` generates a `dist/<category>.d.json.ts` declaration file per token category, so the JSON token modules carry their literal types.

### Migration

Border widths:

| Before | After |
| --- | --- |
| `1`, `'x1'` | `'default'` |
| `2`, `'x2'` | `'medium'` |
| `4`, `'x4'` | `'large'` |
| `8`, `'x8'` | `'extra-large'` |

Border radii:

| Before | After |
| --- | --- |
| `2`, `'2px'`, `'x2'` | `'small'` |
| `4`, `'4px'`, `'x4'` | `'medium'` |
| `8`, `'8px'`, `'x8'` | `'large'` |
| `20`, `'x20'` | `'extra-large'` |
| `9999` | `'full'` |

## Further comments

Internal usages were migrated to the token names with no change to the rendered values: `AudioPlayer`, `TableCell`, `ToastBar`, the `Banner` and `Tabs` stylesheets, and the `Box` and `Message` stories.

`borderWidth()` and `borderRadius()` in `styleTokens.ts` no longer go through `measure()`, since the token lookup covers every accepted value; they also lost the memoization that came with it.

The SCSS `border-width()` and `border-radius()` helpers resolve through `map.get()`, so an unknown token name yields `null` and the declaration is silently dropped. Making them `@error` like `padding()` does would catch typos at compile time, and is worth doing as a follow-up.


dougfabris (approved):
LGTM!

PR #2101, opened 2026-07-17
feat(Table): add cellVerticalAlign prop to TableRow

Pr241singh:

Adds a new `cellVerticalAlign` prop to `TableRow` to allow consumers to apply `vertical-align: top` to all table cells (`td`) within a row.

This provides a reusable API for top-aligning table cells instead of requiring downstream projects to duplicate the styling.

## Issue(s)

Related to RocketChat/Rocket.Chat#41442

## Further comments

This change was extracted into Fuselage following the review feedback on Rocket.Chat PR #41442 so the behavior can be reused by downstream projects.

### DIFF
--- a/packages/fuselage/src/components/Table/Table.styles.scss
+++ b/packages/fuselage/src/components/Table/Table.styles.scss
@@ -79,6 +79,12 @@ $table-selected-border-radius: theme(
       @include clickable;
       background-color: colors.surface(hover);
     }
+
+    &--cell-align-top {
+      .rcx-table__cell {
+        vertical-align: top;
+      }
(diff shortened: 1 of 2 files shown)


dougfabris (changes requested):
use the BEM pattern we have to receive a prop and change the style of the component

Pr241singh:
@dougfabris Thanks for the feedback! I've updated the implementation to follow the existing BEM pattern by introducing a rcx-table__row--cell-align-top modifier and moving the styling into Table.styles.scss. Please let me know if there's anything else you'd like me to adjust.

PR #2138, opened 2026-07-31
feat(fuselage): quieter message header metadata

ivans-netto:
## Summary

[Task](https://rocketchat.atlassian.net/jira/software/c/projects/DSN/boards/2757?selectedIssue=DSN-199)

Part of the **user information redesign** in RocketChat/Rocket.Chat#41650 ([Figma spec](https://www.figma.com/design/CcKhUgDMecW1xvlAtV86o1/Design-system---Web---desktop?m=auto&node-id=18249-35439)), where these styles are currently applied as `.rcx-*` overrides inside the app — this PR moves them into the design system so the app can drop the overrides and consume them from fuselage.

- **Author names** (`rcx-message-header__name`, `rcx-message-system__name` and their shared name container) use the `titles-labels` font color token. When the name container acts as a button (`[role='button']`) it shows a pointer cursor and underlines on hover — in the app it is the hover/click trigger for the user card and full profile. Non-interactive name containers are unaffected.
- **Timestamps** (`rcx-message-header__time`, `rcx-message-system__time`) are quieter through color alone: they stay on the `c1` scale they already declared and use the `secondary-info` font color token. No explicit font sizes — the type ramp has no 10px regular step (`micro` is bold), so the size stays on the scale and the color does the work.
- **Contrast on hover:** `secondary-info` meets WCAG AA on the room surface (4.80:1) but not on the hover surface (4.33:1), so the regular-message timestamp steps up to the `default` token while the message is hovered (11.26:1). System messages don't change surface on hover, so no rule is needed there.

### Impact

This intentionally changes the header metadata of every `Message`/`MessageSystem` consumer — it is the design-system side of the approved redesign, and the app PR above stops overriding these styles once this ships.

## Screenshots

**Message header — before / after** (timestamp in `secondary-info`; name uses `titles-labels`):

| Before | After |
|---|---|
| before | after |

**Hover underline on a clickable author name** (new `WithClickableAuthorName` story):

## Testing

- `WithClickableAuthorName` story added — hover the author name to see the pointer cursor and underline (only when `role='button'` is set).
- 28 visual-regression goldens regenerated; 506 tests passing.
- Jest specs and DOM snapshots unaffected (style-only change); lint/stylelint green.


packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -203,7 +203,9 @@ $message-highlight-border-radius: theme(
       @include use-font-scale(c1);
       flex-shrink: 0;
 
-      color: colors.font(default);
+      color: colors.font(secondary-info);
+
+      font-size: to-rem(10);
```
ggazzo: we don't explicitly set font sizes we use font scale.

ggazzo (dismissed):

packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
```diff
@@ -57,6 +59,10 @@ $message-system-background-color-selected: theme(
     flex-shrink: 0;
 
     margin-inline: lengths.margin(2);
+
+    color: colors.font(secondary-info);
```
ggazzo, line 63: I'm pretty sure the last time we changed that color was for accessibility reasons., https://github.com/RocketChat/fuselage/pull/1157/changes#diff-c8f9ea00adde78a8f16bb4dce15fc1d20ee4add028eb2bcdd51bdd8d51995372R179

are we not introducing a regression here right?

ggazzo:
@juliajforesti 🆘  not 100% sure about my reviews

.changeset/quieter-message-header-metadata.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/fuselage': minor
+---
+
+Quieter message header metadata: author names (regular and system messages) use the titles-labels color token and underline on hover when the name container is a button; timestamps are smaller (10px) in the secondary-info color.
```
juliajforesti: pattern of changeset here is the same as PR title
```suggestion
feat(fuselage): quieter message header metadata
```

packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
juliajforesti, line 63: Yes, this change causes a11y violation:

Image

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -215,14 +217,20 @@ $message-highlight-border-radius: theme(
 
       @include templates.focus-state;
       flex-shrink: 0;
+
+      color: colors.font(titles-labels);
```
juliajforesti, line 226: Why set the color here if you are setting it again in line 225?

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -215,14 +217,20 @@ $message-highlight-border-radius: theme(
...
+
+      &[role='button']:hover {
```
juliajforesti: shouldn't the cursor also change?

juliajforesti (changes requested):

packages/fuselage/src/components/Message/MessageSystem/MessageSystem.styles.scss
ivans-netto: Good catch! The contrast ratio is failing on hover. I'll set the color to be font default on hover so we don't overweight visually the entire chat to cover an failure that happens just once hovering. Thanks!

ivans-netto: FYI I've changed the color contrast on message hover, this way we prevent the contrast to fail

packages/fuselage/src/components/Message/Messages.styles.scss
ivans-netto: The two rules look redundant but do different jobs. The underline on :hover is declared on the container, and CSS paints text-decoration with the color of the box that declares it so without the container color the underline would render in the inherited font-default, mismatched with the titles-labels name under it. Declaring it at the container level is also what makes system messages work for free: they share this same MessageNameContainer, but their name child is a different class (.rcx-message-system__name), so an underline declared on the name would need to know both children's classes, while on the container it covers both. The color on __name is what keeps MessageName correct when it's composed standalone, without the container (Rocket.Chat does this in the Discussions/Threads/VideoConf list items, moderation's ContextMessage and SearchSourceResult). Added comments in cf8f0a65 recording this.

packages/fuselage/src/components/Message/Messages.styles.scss
ivans-netto: I've changed to c1 regular instead

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -203,7 +203,14 @@ $message-highlight-border-radius: theme(
       @include use-font-scale(c1);
       flex-shrink: 0;
 
-      color: colors.font(default);
+      color: colors.font(secondary-info);
+
+      // secondary-info does not meet the contrast requirement against the
+      // hover surface, so the timestamp steps up to the default font token
+      // for as long as the message is hovered.
```
juliajforesti: keep comments brief, and only when strictly necessary

```suggestion
      // secondary-info fails contrast against the hover surface.
```

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -215,14 +222,35 @@ $message-highlight-border-radius: theme(
...
 
     &__name {
       @include use-font-scale(h5);
       @include use-with-truncated-text();
       flex-shrink: 1;
 
-      color: colors.font(default);
+      // Not redundant with the container: MessageName is also composed
+      // standalone (list items, previews), where there is no container to
+      // inherit from.
```
juliajforesti: I'd remove this comment - correlation already explained in line 226-227
```suggestion
```

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -215,14 +222,35 @@ $message-highlight-border-radius: theme(
 
       @include templates.focus-state;
       flex-shrink: 0;
+
+      // Same token as the name below, on purpose: text-decoration is painted
+      // with the color of the box that declares it, so the hover underline
+      // takes this color — without it the container inherits the document
+      // default and the underline mismatches the name under it.
```
juliajforesti:
```suggestion
      // This must match __name's color or the
      // underline mismatches the text under it.
```

packages/fuselage/src/components/Message/Messages.styles.scss
```diff
@@ -215,14 +222,35 @@ $message-highlight-border-radius: theme(
...
+      color: colors.font(titles-labels);
+
+      // The interactive affordance travels as one: the pointer announces the
+      // name is actionable before the click, the underline confirms it under
+      // the cursor. Both are gated on the container actually being a button,
+      // and both reach system messages too, since they share this container.
```
juliajforesti: I would remove this
```suggestion
```

juliajforesti (changes requested):
keep comments brief, and only when strictly necessary - it should not be a history of a debate

juliajforesti (approved):

ivans-netto:
@mergifyio queue

ivans-netto:
@mergifyio queue

PR #2140, opened 2026-08-04
feat(figma-sync): generate Figma component library from Storybook

ggazzo:
Generates Fuselage's Figma component library from Storybook, as a replacement for story.to.design that understands our token architecture.

**Draft** — the workflow added here has never run on a runner. Opening as draft so CI exercises it.

## Why build this instead of buying it

`getComputedStyle` resolves `var()` away and leaves only a hex, so any generic DOM-to-Figma converter has to guess which variable to bind — and Fuselage has 11 pairs of tokens that share a value. Reading the *authored* declaration out of the CSSOM instead exposes the whole chain:

```css
var(--rcx-button-primary-background-color,             /* override hook, never declared */
  var(--rcx-color-button-background-primary-default,   /* component token, DECLARED     */
    var(--rcx-color-blue-500, #156FF5)))               /* primitive, never declared     */
```

Walking it to the first **declared** custom property yields the binding the code actually means. 44% of bindings resolve by exact token name; the rest fall back to matching a variable by value, which the extractor reports.

Two things this surfaced about our own tokens, worth a look independently of this PR:

- Primitives are compile-time only — `--rcx-color-blue-500` does not exist at runtime.
- Button's `warning`, `secondary-warning` and `secondary-success` have no `button-*` custom property, and `border-width` / `border-radius` have none anywhere.

## What's in scope

10 components / 98 variants: Button, Tag, Badge, Callout, Banner, Chip, Label, FramedIcon, Divider, Tooltip.

Twelve more were tried and **rejected rather than shipped wrong** — each recorded in the `skipped` array in `components.json` with its reason. They cluster into three gaps: child-node measurement (ProgressBar, CheckBox, RadioButton, ToggleSwitch, CodeSnippet), gradient/shadow paints (Skeleton, Tile), and SVG vectors (StatusBullet).

## How it stays deterministic

The rule is that AI decides once, config records it, and the pipeline replays it. Picking a `rootSelector` or deciding which args are axes is a reasonable thing to hand to a model; the answer is then frozen into `components.json` and reviewed here, and nothing in the hot path decides anything.

| Guarantee | Enforced by | Fails how |
| --- | --- | --- |
| Config well-formed, skips justified | `yarn workspace @rocket.chat/figma-sync check` | non-zero exit |
| `plugin/code.js` matches `src/apply.js` | same | regenerates, then fails |
| `apply.js` keeps its invariants | same | greps code, not comments |
| Spec shape valid | `validateSpec` in `extract.mjs` | exit 1, names the path |
| A component measured wrong | `error`-level finding | exit 1 |
| A measured value changed | committed snapshot diff | exit 2, prints the diff |
| Re-applying is a no-op | `emit-apply.mjs --twice` | `idempotent: false` |
| Reproducible across machines | Chromium pinned via the lockfile | — |

`figma-spec.snapshot.json` is committed, so any change to a measured geometry or token binding lands as a reviewable diff here rather than propagating quietly.

## Applying the spec is not automated, and cannot be

CI generates and publishes the spec; a human applies it. The blocker is authentication, not headlessness — Figma's remote MCP server *does* write to the canvas server-side with no desktop app and no open editor, but it is OAuth-only ("does not support authentication using personal access tokens and this cannot be enabled") and enforces a client allowlist a GitHub Action cannot satisfy. The REST API cannot create or modify nodes at all.

The one part that *could* be fully automated is the token collections, via `POST /v1/files/:key/variables` with a plan access token — but that endpoint needs an Enterprise plan and we are on Organization.

## Open question for review

`plugin/` may be dead weight. Because the MCP path is headless, an engineer can apply the spec without it. The plugin only earns its place if we want **designers** to trigger a sync. Happy to drop it from this PR.

## Next steps

Ordered by what blocks what. Nothing here belongs in this pull request — it is recorded so the follow-up work is not rediscovered later.

**1. Move the Figma file out of drafts, then publish it.** This is the actual blocker: the 15 component sets and 187 variables exist, but a draft cannot be published as a library, so nobody can consume them. Pick the destination before publishing — moving a file between teams *after* publishing breaks the library link in every file that consumed it.

**2. Publish it as `[WIP] Fuselage Sync`.** We are on the Organization plan, where a published library appears in every member's library picker; scoping a library to one team is an Enterprise feature. Publishing never *enables* anything, so no existing file changes and a component only appears for someone who turns the library on by hand — the name is what stops accidental adoption. The convention already exists in this organization next to `[WIP] Design System - React Native App` and `Icon lib test demo [delete me]`.

**3. Code Connect.** Everything in this pull request pushes code → design. Code Connect points a Figma component back at its source, so a designer inspecting a variant sees the real `<Button variant="primary">`. It publishes over the REST API with a plain token — no OAuth, no client allowlist — which makes it the one remaining piece that CI can own end to end.

**4. Decide what happens to the `Skeleton` page in Figma.** Its three variants shipped invisible before the wrapper/gradient check existed and the component is now in `skipped`, but the page is still in the file with nothing driving it.

**5. Measure a Tamagui component before the SCSS files are removed.** The extractor reads the rendered DOM, not `.scss`, so removing SCSS is not automatically a problem — but that only holds if a Tamagui build still emits a stable root selector and a `var()` chain that reaches a declared custom property. Measuring one `Button` on the Tamagui branch decides whether this costs one function or a rethink.

**6. Phase B gaps**, already listed above: child-node measurement (5 components), gradient and shadow (2), SVG and transforms (2).

Not on this list: re-running the extractor by hand to check whether Figma has drifted. CI now comments on the pull request when the spec stops matching the committed snapshot, with the re-sync commands ready — that gap is what let the library fall a component behind.

## Review notes

- `plugin/code.js` is **generated** from `src/apply.js` — don't review it as hand-written, and don't edit it.
- `yarn.lock` changed only to register the new workspace; without it `yarn install --immutable` fails in CI.
- `figma-spec.json` is gitignored; it's a build artifact. The committed snapshot is the reviewable digest.


julio-rocketchat (changes requested):
Just a reminder to pin the actions versions as hashes (Layne pointed it out). Thank you 🙏🏻

PR #2144, opened 2026-08-06
fix(fuselage-tokens): badge level-1 contrast a11y failing

ivans-netto:
## Summary

[Task](https://rocketchat.atlassian.net/jira/software/c/projects/DSN/boards/2757?selectedIssue=DSN-187)

The dark theme's `badge-level-1` (the `secondary` badge background) was `#484C51`, a color that doesn't match anything in the palette. It now uses `#6C737A` — the value of the `n700` primitive, the same source the light theme references. The literal (rather than a `{colors.n700}` reference) is intentional: the dark palette doesn't reference primitives.

| Theme | badge-level-1 | Source |
| --- | --- | --- |
| Light | #6C737A | `{colors.n700}` (unchanged) |
| Dark | #484C51 → #6C737A | aligned with the n700 value |
| High contrast | #2F343D | `{colors.n800}` (unchanged, intentional) |

White text over #6C737A sits at 4.8:1, above the WCAG AA minimum of 4.5:1 for the badge's micro font size.

## Before / After

secondary badge in the dark theme: before with #484C51, after with the n700 value (#6C737A)

## Storybook

- Reworks the `WithValue` story to show every badge variant with its value and a label with the variant name (rebased on top of #2147, so it covers `secondary`, `primary`, `danger`, and `warning`).

## Testing

- Badge unit tests and full visual regression suite passing (504 tests); `WithValue` golden regenerated — goldens capture the light theme, which is unchanged by the token fix
- Changeset included (`patch` on `@rocket.chat/fuselage-tokens`, same message as this PR title)


.changeset/ghost-badge-contrast.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/fuselage': patch
+---
+
+Fixed the contrast of the `ghost` badge variant. It used `pure-white` text over `stroke-dark`, which made it identical to the `secondary` variant in the light theme and unreadable (~2.5:1) in the dark theme. It now uses `font-titles-labels` over `badge-background-level-0`, meeting WCAG AA in all themes.
```
juliajforesti: remember changeset should be the same as PR title

packages/fuselage/src/components/Badge/Badge.styles.scss
juliajforesti, line 1: I'm not sure I follow this decision: are we bringing back this variant? Why? 
Although the code still kept the vars of `ghost`, we moved away from this naming and started using `secondary`. This decision came from the design team - even in Figma there is no `ghost` option:
Image

That's why it was "failing" dark theme - there was never such support.

Why not just use the `secondary` variant?

juliajforesti (changes requested):

packages/fuselage/src/components/Badge/Badge.styles.scss
ivans-netto: That's true. It's really hard to keep track of these things, if these components shouldn't be used, they should exist, right? 

The case here is because we wanna use this in the workspace tabs (electron) and secondary offers a low contrast against the surface applied there. 

I'll try to keep our decision tho, I'll improve the contrast in the secondary (which is kinda an issue in other places now) and delete the ghost.

packages/fuselage-tokens/src/badge/base.json
```diff
@@ -16,7 +16,7 @@
     },
     "dark": {
       "level-0": { "value": "#404754" },
-      "level-1": { "value": "#484C51" },
+      "level-1": { "value": "{colors.n700}" },
```
juliajforesti: level-1 has the same color in light and dark theme? is this correct? 
also, this is the only token that references the primary palette... idk if this is good

juliajforesti (dismissed):

packages/fuselage-tokens/src/badge/base.json
ivans-netto: yes, it's correct... they worked well this way. 
regarding the primitive, I could set the hex vale directly to keep the dark theme approach

.changeset/badge-level-1-n700.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rocket.chat/fuselage-tokens': patch
+---
+
+fix(fuselage-tokens): align dark badge level-1 background with the n700 value
```
juliajforesti:
```suggestion
fix(fuselage-tokens): badge level-1 contrast a11y failing
```

juliajforesti (changes requested):

juliajforesti (approved):

PR #2154, opened 2026-08-11
chore(mp3-encoder)!: Remove the `@rocket.chat/mp3-encoder` package

tassoevan:

Removes the `@rocket.chat/mp3-encoder` package from this repository.

Fuselage is a design system repository, and a LAME MP3 encoder has nothing to do with a design system. The package had a single consumer — the audio message recorder in Rocket.Chat — and living here meant every change to it required a release cycle across two repositories before it could reach the product. The code has been migrated to Rocket.Chat, next to its only consumer, in RocketChat/Rocket.Chat#41745.

What this PR does:

- Deletes `packages/mp3-encoder` in full (sources, tests, WAV fixtures, build config).
- Drops the package row from the root `README.md`.
- Removes the `new-cap: 'off'` override for `packages/mp3-encoder/**/*.ts` from `eslint.config.mjs` — it existed only because the LAME sources mirror the upstream C identifiers, and that exemption now lives in the Rocket.Chat ESLint config.
- Regenerates `yarn.lock`.

No changeset: the package is gone, so there is nothing left to version here. Nothing else in this repository imports it.

## Issue(s)

[ARCH-2358](https://rocketchat.atlassian.net/browse/ARCH-2358)

## Further comments

**Blocked by RocketChat/Rocket.Chat#41745.** That PR is the other half of the move: it brings the sources into Rocket.Chat and switches `apps/meteor` from `@rocket.chat/mp3-encoder: ^0.31.26` to `workspace:~`. Until it merges, deleting the package here would leave the encoder without a home, so this should not land first.

Removing the package from the monorepo does not remove it from npm. `@rocket.chat/mp3-encoder@0.31.26` stays published and installable, so any external consumer keeps working; the migrated package in Rocket.Chat keeps the same name and version, so the npm line continues from there. A separate `npm deprecate` pointing at the new location is the follow-up, and it is worth doing once #41745 is released.

The last released version from this repository is `0.31.26` — the same version the migrated package starts from, so no version is skipped or reused.

[ARCH-2358]: https://rocketchat.atlassian.net/browse/ARCH-2358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gabriellsh (approved):
See you around, econder! Hope you had a nice stay...

PR #2162, opened 2026-08-13
feat(fuselage): Update input pseudo states

abhinavkrin:

Updates the pseudo states of the input components to match the design spec.

- Placeholder text now uses `font-hint` and value text `font-titles-labels`. The placeholder previously used `font-annotation`, which is ~2.6:1 against white and fails WCAG AA; `font-hint` is ~4.7:1 and passes. The value colour was darkened alongside it to keep the contrast relationship.
- Disabled inputs use `font-disabled` instead of `font-default`, so disabled text is now visually distinct from enabled text.
- Invalid inputs no longer tint their value text red; only the border stays red, per the design file.
- Icons inside inputs follow the input's font colour in every state, including focus, where they were previously blue.
- Adds a hover state. The `on-hover` hook already existed in the shared input primitive but resolved to the same colour as the resting border, so hovering was a no-op. It now resolves to `stroke-dark`.
- Every input type now shares the same transition. The 0.18s fade comes from the `animated` Box prop, and only `InputBox` passed it, so `Select`, `MultiSelect`, `AutoComplete` and `PaginatedSelect` snapped between colours instantly. They inherit their border and colour tokens from `%rcx-input-box` through `@extend`, but `@extend` cannot reach `.rcx-box--animated` because that class is applied from JSX. Passing `animated` on the five outer boxes that carry the input border means all of them share one definition, including its `prefers-reduced-motion` guard, rather than a duplicated timing value.

Two related fixes found while working through the states matrix:

- A pressed input no longer loses its state. `active-border-color` used the lighter `stroke-medium`, so pressing made the border lighter than hover, and an invalid input dropped its red border entirely while pressed.
- The `::placeholder` rule in `InputBox/mixins.scss` had been commented out behind a "fix placeholder performance issue" TODO since 2019, so native placeholders rendered in the browser's default colour and no placeholder token ever reached them. It is re-enabled; under plain SCSS it compiles to 8 rules totalling ~2.3KB.

Because `InputBox` is the base for the whole input family, the token changes apply to Select, MultiSelect, AutoComplete and PaginatedSelect as well. Those components set some colours from JSX props rather than the shared mixins, so five hardcoded `default` values were swapped to `titles-labels` to keep them consistent.

## Issue(s)

CORE-2375

## Further comments

A few token pairs now resolve to the same value. These are intentional, not oversights:

- `focus-icon-color` equals the base text colour, because the ticket asks that icons follow the font colour "always" and the design file shows a dark icon in the focus row.
- `hover-border-color` and `active-border-color` are both `stroke-dark`, and `invalid-active-border-color` matches the invalid border. The design file has no pressed state, and the alternative was letting a press downgrade the state the user is already in.

Known and deliberately out of scope, both pre-existing:

- A disabled `Select`/`AutoComplete` does not dim its value text. `Box`'s `color` prop emits `!important`, so the disabled token cannot win by specificity; fixing it means removing those props and wiring the tokens through `.rcx-select`. Worth its own PR.
- `with-icon-addon-colors` on `.rcx-select` emits sibling selectors while the addon is a descendant, so it never matches, and `.rcx-autocomplete__addon` has no rule at all.

Because `transition: all` now applies to the Select family's outer box, a size change animates too. A `MultiSelect` growing as chips are added is the visible case. `InputBox` has always behaved this way, so this is the consistency the ticket asks for rather than a new effect, but it is the one change that goes beyond colour.

Visual regression: run locally in the Playwright Docker image. 15 baselines changed, all of them input stories, and they are committed here. Every diff is colour-only, with identical image dimensions before and after, so there is no layout shift. The full suite is green at 506 passed. The transition change adds no further baseline churn, since screenshots are taken with `animations: 'disabled'` and a transition only affects intermediate frames.

Downstream note: the equivalent placeholder colour in Rocket.Chat's `MessageComposerInput` is hardcoded to `font-annotation` and should move to `font-hint` when the version is bumped.

[CORE-2375](https://rocketchat.atlassian.net/browse/CORE-2375)

[CORE-2375]: https://rocketchat.atlassian.net/browse/CORE-2375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DSN-200](https://rocketchat.atlassian.net/jira/software/c/projects/DSN/boards/2757?selectedIssue=DSN-200)


packages/fuselage/test/snapshots/inputs-passwordinput--disabled-desktop.png
juliajforesti, line 1: this snapshot is showing some size change, can you check if it's expected @abhinavkrin ?

Image

juliajforesti (changes requested):

packages/fuselage/test/snapshots/inputs-passwordinput--disabled-desktop.png
abhinavkrin: I looked into this and I don't think it's coming from this PR, though it's worth a second pair of eyes.

What I can show: I rebuilt the code from before my branch and it already renders this input at 38px, so the committed 40px baseline looks like it was already out of date. Comparing that baseline against the old render using Playwright's default 0.2 threshold gives 0 differing pixels (at 0.1 it's 2488), which would explain why this never failed a test before.

My reading is that the lighter icon colour is what finally pushed the diff over the threshold, so `--update-snapshots` rewrote the file with the current geometry, but I haven't isolated that part and I also haven't tracked down when the 40px baseline stopped matching.

Side by side, the baseline regenerated from the code before my branch vs the one in this PR, both 38px with identical box edges:

pw-side-by-side

juliajforesti (approved):

PR #2164, opened 2026-08-14
feat(fuselage): Add `joined` ButtonGroup variant with `ghost` button segments

ivans-netto:

Implements the "joined" ButtonGroup variant from the design-system Figma exploration (generalized from the video-call joined icon button):

- New `joined` prop on `ButtonGroup` (`rcx-button-group--joined`): fuses the buttons into a single segmented control — zero gap, group-level border radius (same token as Button, 0.25rem) with hidden overflow, children flattened to `border-radius: 0`, no dividers.
- The joined group carries a translucent background: the secondary button background token at 60% via `color-mix` — it theme-switches with the token and never fades the children.
- New `ghostPosition?: 'start' | 'end'` prop on `ButtonGroup` (design discussion with @juliajforesti): the **group** picks the ghost segment, turning its first or last button fully transparent so the group background (and the surface behind it) shows through. Styled purely via edge selectors (`.rcx-button-group--ghost-start > .rcx-button:first-child`), so a middle segment can't be ghosted by design; the prop type is a union that only accepts `ghostPosition` together with `joined: true`, so misuse is a compile-time error — no runtime warnings. `Button`/`IconButton` have no ghost API. Hover/active/focus feedback uses the secondary state tokens at 60%; label/icon color keeps the on-secondary font token.
- `joined` takes precedence over `small`/`large` spacing and composes with `vertical`, `stretch`, and `align` (joined groups hug their content; `align` is reproduced with auto margins, `stretch` restores full width, vertical stacks keep segments flush).
- Stories: `Joined` (confirmation pair + video-call device controls with ghost segments), `JoinedSegmented`, `JoinedSplitButton`, `JoinedVertical`, `WithIconButtons`, and `JoinedWithMultipleButtonSizes` (any size works as long as the joined group is homogeneous) — joined stories sit on a surface-light wrapper so the ghost translucency reads in dark mode, and the gap-only `small`/`large` controls are hidden there. Docs page updated with joined + ghost usage rules (2–4 buttons, at most one primary, don't mix regular Buttons with IconButtons, keep sizes homogeneous).

No breaking changes to the existing ButtonGroup API.

### Why every ButtonGroup story is now snapshot-tested

Most of this diff (~1k lines) is generated test snapshots, not product code. The spec was upgraded to the repo's snapshot-per-story convention (as in `Tag`, `StatusBullet`, etc.), which turns the Storybook itself into the component's executable contract:

- Every story on the docs page is rendered in CI with a DOM snapshot **and** an axe accessibility check — the Storybook can't silently drift from what ships.
- Adding a story is adding coverage: any future variant documented in Storybook is automatically render- and a11y-tested, with no extra test code.
- The 12 pre-existing ButtonGroup stories gained render/a11y coverage they never had; the visual-regression baselines complete the picture, making Storybook the single source of truth for the component's visual and accessibility spec.

### Screenshots

| Story | Preview |
| --- | --- |
| Joined (confirmation pair + video-call device controls with ghost segments) | Joined |
| JoinedSegmented | JoinedSegmented |
| JoinedSplitButton (ghost chevron segment) | JoinedSplitButton |
| JoinedVertical | JoinedVertical |
| WithIconButtons (default gap vs joined) | WithIconButtons |

Task: [DSN-195](https://rocketchat.atlassian.net/jira/software/c/projects/DSN/boards/2757?selectedIssue=DSN-195)
Reference: [Figma — Design system / Button group](https://www.figma.com/design/CcKhUgDMecW1xvlAtV86o1/branch/Dghwk1KqJJ3DCGwUw5XpRk/Design-system---Web---desktop?node-id=31082-26)

[DSN-195]: https://rocketchat.atlassian.net/browse/DSN-195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/fuselage/src/components/ButtonGroup/ButtonGroup.styles.scss
```diff
@@ -47,4 +50,81 @@
...
+    .rcx-button--ghost {
+      color: colors.button(on-secondary);
+      border-color: transparent;
+      background-color: transparent;
+
+      @include on-focus-visible {
+        border-color: transparent;
+        background-color: color-mix(
+          in srgb,
+          #{colors.button(secondary-focus)} 60%,
```
juliajforesti: No need for `#{}` here
```suggestion
          colors.button(secondary-default) 60%,
```

packages/fuselage/src/components/ButtonGroup/ButtonGroup.styles.scss
```diff
@@ -47,4 +50,81 @@
...
+          transparent
+        );
+      }
+
+      @include on-hover {
+        border-color: transparent;
+        background-color: color-mix(
+          in srgb,
+          #{colors.button(secondary-hover)} 60%,
```
juliajforesti: same here

packages/fuselage/src/components/ButtonGroup/ButtonGroup.styles.scss
```diff
@@ -47,4 +50,81 @@
...
+          transparent
+        );
+      }
+
+      @include on-active {
+        border-color: transparent;
+        background-color: color-mix(
+          in srgb,
+          #{colors.button(secondary-press)} 60%,
```
juliajforesti: same here

packages/fuselage/src/components/ButtonGroup/ButtonGroup.styles.scss
```diff
@@ -47,4 +50,81 @@
...
+      border-color: transparent;
+      background-color: transparent;
+
+      @include on-focus-visible {
+        border-color: transparent;
+        background-color: color-mix(
+          in srgb,
+          #{colors.button(secondary-focus)} 60%,
+          transparent
+        );
```
juliajforesti: I'd suggest creating a css variable in `packages/fuselage/src/styles/variables/button-colors.scss` with this color-mix, since it is repeating 3 times in this file

juliajforesti (changes requested):
On a more broad level review, this new `ghost` prop is exposed as a Button/IconButton-level prop, but it only has an effect when the button is a descendant of `ButtonGroup joined`. But nothing **enforces** that relationship - the only clue is the JSDoc comment.
So I'm trying to think about other approaches to this. We need to discuss this a bit further...

packages/fuselage/src/components/ButtonGroup/ButtonGroup.styles.scss
ivans-netto: extracted the repeated color-mix into themeable variables in button-colors.scss ($button-group-joined-background-color + the three ghost state variables), built on a private -translucent() helper so the 60% recipe lives in one place

ivans-netto:
> On a more broad level review, this new `ghost` prop is exposed as a Button/IconButton-level prop, but it only has an effect when the button is a descendant of `ButtonGroup joined`. But nothing **enforces** that relationship - the only clue is the JSDoc comment. So I'm trying to think about other approaches to this. We need to discuss this a bit further...

You're right, before, nothing stopped someone from using ghost on a button outside a joined group.

Fixed in 41403cc4. Now ButtonGroup tells its children (via an internal React context) whether it is joined. Button and IconButton check that: inside a joined group, ghost works; anywhere else, it does nothing and logs a dev-only warning. So the relationship is enforced at runtime, not just documented.

The CSS was already scoped under --joined, so that stays as a safety net. Added unit tests for the three cases (inside joined, inside a normal group, outside any group). The context is internal, the public API doesn't change.

If you'd still prefer not having the prop on Button at all, another option is a compound component like ButtonGroup.Ghost. Happy to switch to that if you like it better.

@juliajforesti

juliajforesti:
> > On a more broad level review, this new `ghost` prop is exposed as a Button/IconButton-level prop, but it only has an effect when the button is a descendant of `ButtonGroup joined`. But nothing **enforces** that relationship - the only clue is the JSDoc comment. So I'm trying to think about other approaches to this. We need to discuss this a bit further...
> 
> You're right, before, nothing stopped someone from using ghost on a button outside a joined group.
> 
> Fixed in [41403cc](https://github.com/RocketChat/fuselage/commit/41403cc4c726837114ca8e280034a31b87105b54). Now ButtonGroup tells its children (via an internal React context) whether it is joined. Button and IconButton check that: inside a joined group, ghost works; anywhere else, it does nothing and logs a dev-only warning. So the relationship is enforced at runtime, not just documented.
> 
> The CSS was already scoped under --joined, so that stays as a safety net. Added unit tests for the three cases (inside joined, inside a normal group, outside any group). The context is internal, the public API doesn't change.
> 
> If you'd still prefer not having the prop on Button at all, another option is a compound component like ButtonGroup.Ghost. Happy to switch to that if you like it better.
> 
> @juliajforesti

I'll push an approach so we can evaluate.
But other than that, are we only supporting one sizing? adding the size props to the buttons inside the joined group has zero effect - which is also weird...

juliajforesti (approved):

juliajforesti (approved):

PR #2191, opened 2026-09-09
chore(fuselage): Overhaul the build and move it to Vite

tassoevan:
Housekeeping on `@rocket.chat/fuselage`'s build, which was the last package still on webpack. Eleven commits, each independently verified — every one of them was checked by rebuilding and diffing the artifacts rather than by inspection.

## Fixes that reach consumers

**Direction-aware CSS now reaches the browser.** `postcss-logical` was configured as `{ preserve: true }`, but `preserve` is not an option in v9 — it was silently ignored, so the plugin ran at its defaults and compiled all 186 logical properties in the sources down to hardcoded left-to-right physical ones. Exactly one survived into `dist/fuselage.css`. `postcss-dir-pseudo-class` was separately replacing `:dir(rtl)` with the weaker `[dir="rtl"]` ancestor selector, which only matches under an explicit `dir` attribute and misses `dir="auto"`. Both plugins are gone and the CSS ships as authored. 181 rules change; **LTR rendering is identical** and all 499 visual-regression snapshots pass untouched — RTL is where the behaviour changes, and that is the fix.

**Development and production stylesheets no longer collide.** Both webpack modes emitted `dist/fuselage.css`, so the production run overwrote the development one — but with source maps off in production, nothing overwrote `dist/fuselage.css.map`, leaving a map describing a file that was gone. `dist/fuselage.css` is unchanged; development now emits `dist/fuselage.development.css`.

## Build simplification

- **Babel removed entirely.** It was configured in two places and did nothing in either: the `/\.js$/` rule never matched (the sources are 467 `.tsx`, 122 `.ts`, 87 `.scss`, no `.js` at all) and the `babel-loader` in the SCSS chain only transpiled the JS stub `css-loader` emits, which `MiniCssExtractPlugin` lifts the stylesheet out of anyway. `.babelrc.js` also claimed IE 11 and Safari 7 targets while `ts-loader` actually emits ES2024. Drops 96 packages (−1.5 MiB).
- **Three redundant `.styles.scss` side-effect imports dropped** — already reachable through `Messages.styles.scss`; `dist/fuselage.css` byte-identical.
- **The webpack config moved to TypeScript** and is now type-checked by `lint`, which it was not before: it carried a `@type {import('webpack').Configuration}` annotation that nothing enforced. No new tooling — Node strips types natively and webpack-cli already discovers `.mts`.
- **`@rocket.chat/fuselage-tokens`' unused stylelint setup removed.** Its `.stylelintrc` was the only reason the package pinned `postcss`, and nothing ever ran it: no `lint` script, no invocation in `build.js`, and no lintable stylesheet — the sources are JSON.

## The bundler swap

The package now builds with **Vite** instead of webpack. Same Rollup underneath as the rest of our tooling, and the reason the CSS pipeline could move at all is that the postcss chain had already shrunk to `autoprefixer` + `cssnano`.

| artifact | webpack | Vite |
|---|---:|---:|
| `fuselage.production.js` | 216 KB | **126 KB** |
| ↳ gzipped | 44 KB | **32 KB** |
| `fuselage.development.js` | 613 KB | **271 KB** |
| `fuselage.css` gzipped | 24,101 | **24,101** |

Equivalence was measured, not assumed: both bundles export the **same 282 names with the same types**, 15 components render byte-identical markup from either build, and all 7 Safari-critical `-webkit-` prefixes survive — that last one matters because the visual-regression suite renders in Chromium and would pass with them missing.

**Known cost:** `dist/fuselage.development.css.map` is no longer published. Vite emits no source map for extracted CSS in a build, and restoring it would mean compiling sass and postcss a second time purely to produce one — a map from a separate compilation risks not describing the CSS Vite actually emitted, which is worse than none. This partially undoes the stylesheet-split fix above, and it is recorded in the changeset rather than hidden.

## Review notes

- Three changesets, all `patch`; `fuselage.api.md` is unchanged across all eleven commits. There is an argument for `minor` on the Vite one — a 42% bundle change is more than consumers usually expect from a patch.
- **webpack does not leave the package**: Storybook is still on `@storybook/react-webpack5` and its loader chain needs css-loader, postcss-loader, resolve-url-loader, sass-loader and style-loader. Only `ts-loader`, `wrapper-webpack-plugin` and `mini-css-extract-plugin` become removable.
- `autoprefixer` was deliberately **kept**. Removing it looked tempting but Safari — including 26.5 and Technology Preview — still reports `user-select` and the `text-decoration` shorthand as prefix-only in the `caniuse-lite` version this package pins.


packages/fuselage/webpack.config.mts
```diff
@@ -0,0 +1,109 @@
+import { resolve } from 'node:path';
```
ggazzo: 👀

ggazzo (approved):

PR #2212, opened 2026-09-18
feat(fuselage): Add a `badge` prop to `IconButton`

ggazzo:
`IconButton` gains a `badge` prop: what to pin to the button's corner.

The pattern already lived here, as a story people copied out — a relatively positioned `IconButton`, an absolutely positioned `Box` and a `Badge` inside it. In Rocket.Chat alone four copies of it had grown across three files, and they had drifted apart in how far the badge overhung and how it was announced. It is a prop now, so the placement, the overhang and the RTL flip come from one place.

```tsx
<IconButton
  icon='balloon'
  aria-label={t('Chat__count__unread', { count: unread })}
  badge={unread > 0 ? <Badge variant='danger'>{unread}</Badge> : undefined}
/>
```

- `badge` is any `ReactNode`, so the caller decides variant, title, and whether it is a count or the childless dot. No badge is no corner, not an empty one — nothing is rendered and the button keeps its usual `overflow`.
- The badge is hidden from assistive technology, always: `aria-label` replaces a button's contents rather than adding to them, so a badge inside a labelled button is never read out. What it says has to reach the button's own name, which is the caller's to write — the old `role='status'` in the stories promised an announcement that never happened.
- `.rcx-button__badge` places it with logical insets, and `:dir(rtl)` flips the overhang; the copied story used `top`/`right`, which put the badge inside the button in RTL.
- `_IconButtonWithBadge` and `_IconButtonAvatarWithBadge` now use the prop instead of hand-rolling the position.

Follow-up: Rocket.Chat has a local wrapper for this (RocketChat/Rocket.Chat#41956) that drops once this is released.


packages/fuselage/src/components/Button/IconButton.tsx
```diff
@@ -117,13 +128,22 @@ function IconButton({
...
       {...props}
     >
       {isValidElement<any>(icon) ? (
         icon
       ) : (
         <Icon name={icon} size={getIconSize()} />
       )}
+      {hasBadge && (
+        // Hit-testable, because a badge carrying a `title` promises a tooltip and `pointer-events: none` is
+        // precisely what stops one appearing. The badge sits inside the button, so a click on it is a click on
```
dougfabris: We can live without this comments

packages/fuselage/src/components/Button/IconButton.tsx
```diff
@@ -117,13 +128,22 @@ function IconButton({
       {...kindAndVariantProps}
       {...getSizeClass()}
       rcx-button--icon-pressed={pressed}
+      rcx-button--with-badge={hasBadge}
```
dougfabris: Just use "badge"

packages/fuselage/src/components/Button/IconButton.tsx
```diff
@@ -15,6 +15,14 @@ export type IconButtonSize = {
 
 export type IconButtonProps = {
   icon: IconName | ReactElement<any>;
+  /**
```
dougfabris, line 18: These comments are not necessary

dougfabris (changes requested):

packages/fuselage/src/components/Button/IconButton.tsx
ggazzo: I this one is kinda, just reduced the verbosity

dougfabris (approved):

